PR #172, opened 2015-08-09
Adding Varnish Into Project

bradallenfisher:
I ran this 3 times on my VM. 

yes | yum -y install git && git clone https://github.com/bradallenfisher/elmsln.git /var/www/elmsln && bash /var/www/elmsln/scripts/install/handsfree/centos/centos-install.sh someurl ln someurl.com http myemail@gmail.com yes

It will need some more varnish vcl love in the future, but let's do some benchmarking and see what you are looking for it to acheive.

### DIFF
--- a/docs/domains.txt
+++ b/docs/domains.txt
@@ -1,41 +1,41 @@
 #ELMSLN domains.conf recommendations
-NameVirtualHost *:8080
-<VirtualHost *:8080>
+NameVirtualHost *:80
+<VirtualHost *:80>
     DocumentRoot /var/www/elmsln/domains/online
     ServerName online.YOURUNIT.edu
     ServerAlias DATA.online.SERVICEYOURUNIT.edu
 </VirtualHost>
-<VirtualHost *:8080>
+<VirtualHost *:80>
     DocumentRoot /var/www/elmsln/domains/courses
     ServerName courses.YOURUNIT.edu
     ServerAlias DATA.courses.SERVICEYOURUNIT.edu
 </VirtualHost>
-<VirtualHost *:8080>
+<VirtualHost *:80>
     DocumentRoot /var/www/elmsln/domains/media
     ServerName media.YOURUNIT.edu
     ServerAlias DATA.media.SERVICEYOURUNIT.edu
 </VirtualHost>
-<VirtualHost *:8080>
+<VirtualHost *:80>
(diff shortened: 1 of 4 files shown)


bradallenfisher:
Ummm... this may have broken things even though it comes online. I am not familiar enough to know what is supposed to work and how. I would recommend running the one liner script using my fork and making sure that the product is working before merging.

btopro:
Make sure online,media,comply address comes up. other addresses are brought up after CIS tells them to be created, like courses domain

Wp8

bradallenfisher:
They come up. The one thing I noticed is that on sing 100 i can't get System Setup to do anything. It just blinks at me. There might be something the VCL is doing to the progress.js or maybe something else even.

btopro:
Hmm ok, ill apply this to my one line and see on a diff host to test

Wp8

bradallenfisher:
Yeah Travis came back OK.  Like server side stuff is good.  Just need to
make sure the application is OK.

btopro:
I think Travis is a false positive on this BC I haven't accounted for pull requests when I spin things up so its pulling from master main repo I think, can check in yml for it

Wp8

btopro:
Nah Travis looks like its reading off  correct PR and all spun up there. Will need to try in DO and look deeper

bradallenfisher:
https://github.com/bradallenfisher/elmsln/blob/master/docs/varnish.txt this file has the VCL in it... 
If when you spin it up it isn't doing what you want I will strip out all the request handling and see if it works from there... then we can incrementally add new language for specific hits and misses.

btopro:
Haven't tested yet but just in looking through the commit...
Can you not modify the domains files included and instead sed it to the correct state? There are other places that modify this file and your commit would currently bork the non centos installation routine(s)

bradallenfisher:
Gotcha. Yes.

bradallenfisher:
looks like elmnsln.conf is created after scripts/install/handsfree/centos/centos-install.sh runs so sed can't do it from here. suggestions?

bradallenfisher:
i will initiate another pull.

btopro:
make sure we get chkconfig varnish on in there too so that server restarts varnish when it reboots

bradallenfisher:
here comes another pull

PR #174, opened 2015-08-09
From phone...lol

bradallenfisher:
At my birthday party.


btopro:
testing manually, this wouldn't affect travis since travis is a custom install routine based off of ubuntu (which this also currently only supports centos)

bradallenfisher:
did you determine if this will bork the app in any way?

btopro:
From what I could tell no, I was able to setup a course space as I could before; Varnish won't kick in for the webservice calls since they have auth cookies so it really should just be a gain

Issue #80, opened 2015-05-28
Create a ELMSLN system level user during install


btopro:
provide bash script to do this as a stand alone as well. Then, installation would tap that user instead of root on the crontab (for instance). This is similar to what AEgir does for added security in that the elmsln user would have ownership of just the elmsln directory and write access on the files its supposed to. Again, getting more OS-like with this component.

btopro:
marking security in that it is security related, not a security bug by any means.

bradallenfisher:
assign me

btopro:
We'll talk today but please work on this via fork and PR

bradallenfisher:
do you still want the root user to run https://github.com/btopro/elmsln/blob/master/scripts/install/root/elmsln-preinstall.sh

btopro:
This script to create the elmsln user for later use in running calls would be the thing that's called and established as part of elmsln-preinstall so, basically yes, root would need to run to initially kick off the entire process but then after taht I want elmsln being the account that handles the rest of the calls / anyone in the elmsln group.

bradallenfisher:
Gotcha. What are the commands you want elmnsln  to run without passwd using sudo.

btopro:
which ones are worth doing it for? lol, never fully understood that cause it seems like if you had chown / chmod you could just modify something to be your ownership and do whatever you want

bradallenfisher:
Thinking more like restarting services like httpd mysql etc. Any other system level operation. I will come up with a list and go from there.

btopro:
technically would be nice if it could update php / apache / mysql settings (so would need to be able to reboot them).

There's also some bash scripts in the scripts folder it would be good to be able to run (basically anything that's not specifically the root system installer). Can it get that granular as to individual scripts?

bradallenfisher:
yeah on install we could just set the perms to be 700 and make the owner elmsln for any scripts that can only be run by the elmnsln user.

There is a lot going on in the scripts directory. I have forked the project here https://github.com/bradallenfisher/elmsln and will add the script...

How can i test this in an environment? 
Should i follow https://github.com/btopro/elmsln/blob/master/INSTALL.txt
What script(s) in order do I run to set this up in a completely blank centos env?

btopro:
https://drupal.psu.edu/blog/post/install-elmsln-digital-ocean-one-line or https://github.com/btopro/elmsln-vagrant

bradallenfisher:
[...]

PR #227, opened 2015-08-30
Went back and forth on some stuff

bradallenfisher:
Basically this adds the new VM from puphpet, which breaks the tty thing for sudo. Rather it didn't correct it like the chef version did. So I am removing the required tty for Vagrant only. It probably isn't necessary to require tty anymore according to redhat's docs but i didn't want to assume anything. If you want that we can move the inline call into handsfree centos. 

As a bonus this vm out of the box handles RVM as a global... so the user on the system would just need to be added to the group rvm. Again i am not assuming "wants", though it would be real easy to add the vagrant user to the rvm group and you could bypass all need to dl and setup ruby on install.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -42,19 +42,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     # you can modify these manually if you want specific specs
     v.customize ["modifyvm", :id, "--memory", mem]
     v.customize ["modifyvm", :id, "--cpus", cpus]
-
   end
-
-
+  
   # run script as root
   config.vm.provision "shell",
     path: "scripts/vagrant/handsfree-vagrant.sh"
-  # run inline script as root
 
-#remove the error sudo: sorry, you must have a tty to run sudo
-#see //github.com/mitchellh/vagrant-rackspace#centos--rhel--fedora
-  config.vm.provision "shell",
-    inline: "sed -i 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers"
+  # run inline script as root
+  #remove the error sudo: sorry, you must have a tty to run sudo
+  #see //github.com/mitchellh/vagrant-rackspace#centos--rhel--fedora
+    config.vm.provision "shell",
+      inline: "sed -i 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers"
 
   # run as the vagrant user
   config.vm.provision "shell",
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -16,6 +16,35 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.network "forwarded_port", guest: 80, host: 80
   config.vm.network "forwarded_port", guest: 3306, host: 3306
 
+  # automatically carve out 1/4 of the box resources for this VM
+  config.vm.provider "virtualbox" do |v|
+    host = RbConfig::CONFIG['host_os']
+
+    # Give VM 1/4 system memory & access to all cpu cores on the host
+    if host =~ /darwin/
+      cpus = `sysctl -n hw.ncpu`.to_i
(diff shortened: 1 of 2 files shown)


btopro:
errr. what is this fixing exactly?

bradallenfisher:
The fact that chef/centos doesn't exist anymore.

btopro:
ahhhh ok, I thought you were saying all of vagrant cloud hosted images were password protected

bradallenfisher:
Yeah. Some are private now.  Some aren't.  It may be wise at some point to create anew elmsln hosted box so we don't have this problem everytime some one marks their box private.

btopro:
yeah we used to have one but switched to chef hoping that ppl might already have that box instead of getting one fresh but your right; might make sense to snag an image and stuff it back up there in vagrant cloud (i was hosting one out of box for awhile). We could also apply the 5.5 php / mysql upgrades to it ahead of time so that once people have it once it's faster to spin up afterwards, just a thought. the pre-install routine is why I wasn't doing that anymore but that hasn't changed in about a year now

btopro:
". Again i am not assuming "wants", though it would be real easy to add the vagrant user to the rvm group and you could bypass all need to dl and setup ruby on install." 

So are you saying we could simplify / get rid of the vagrant installer somehow and just go w/ a normal centos install? Vagrant install is VERY similar but yet slightly different bc of the issue you mentioned w/ the TTY

bradallenfisher:
no sorry. I am saying that the puphpet version of centos has done some up
front work with installing ruby.
In our scripts we wouldn't need to try to install the gpg keys and use the
/curl thing for ruby/rvm.

All we would need to do is add the vagrant user to the rvm group.

usermod -a -G rvm vagrant  (example
https://github.com/bradallenfisher/drupal-dev-vagrant/edit/master/scripts/installer/root.sh
)

My not assuming stuff is that I don't know if everyone cares about front
end stuff... though it probably wouldn't hurt anything to add vagrant to a
group.

On Sun, Aug 30, 2015 at 3:27 PM, Bryan Ollendyke <notifications@github.com>
wrote:

> ". Again i am not assuming "wants", though it would be real easy to add
> the vagrant user to the rvm group and you could bypass all need to dl and
> setup ruby on install."
>
> So are you saying we could simplify / get rid of the vagrant installer
> somehow and just go w/ a normal centos install? Vagrant install is VERY
> similar but yet slightly different bc of the issue you mentioned w/ the TTY
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/elmsln/elmsln/pull/227#issuecomment-136176398>.
>



-- 
Develop for the future.
bradallenfisher on github <https://github.com/bradallenfisher>

btopro:
I mean it's vagrant and I don't tell people to run off the Vagrant copy for production (though we try to stick close to that).  What if Vagrant was added to that group and #80 was produced during install to install the system? This would be pretty secure I'd think, something that could even be run in prod (assuming base-box wasn't used and this was instead deployed against remotes)

bradallenfisher:
it may be moot now... for some reason vagrant can't run sudo commands on
the box. :(

On Sun, Aug 30, 2015 at 3:35 PM, Bryan Ollendyke <notifications@github.com>
wrote:

> I mean it's vagrant and I don't tell people to run off the Vagrant copy
> for production (though we try to stick close to that). What if Vagrant was
> added to that group and #80 <https://github.com/elmsln/elmsln/issues/80>
> was produced during install to install the system? This would be pretty
> secure I'd think, something that could even be run in prod (assuming
> base-box wasn't used and this was instead deployed against remotes)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/elmsln/elmsln/pull/227#issuecomment-136177428>.
>



-- 
Develop for the future.
bradallenfisher on github <https://github.com/bradallenfisher>

bradallenfisher:
closing this cause I have a new box now with the right settings for sudo stuff.

Issue #206, opened 2015-08-21
Content styling / widget guide


btopro:
We've tried for years to do this and now I think we might have the right person to really make it a thing.

We need a content style guide for describing instructional components. We've got a lot of thought into how this starts to happen already but have yet to officially start doing it. Things involved in this:
- [x] Add different heading options to the styleguide #307
- [x] templates that match these attributes
- [x] instructional significance of components aligned w/ visual significance
- [x] Content Layout Ideas to Blueprint/Template #1074

- [x] accessibility minded default selections and templates
- [ ] definition of different attributes that can have color associated with them
- [ ] Documentation / Official style guide page that illustrates all supported components (and that can serve for visual regression)

As we're on Materialize this gets a lot easier to build new templates / have design criteria to build off of. See http://materializecss.com/ for html you can use and we've started adding some templates in that match this styling already.

btopro:
Related: #42 #205 #155

rickhumphries:
Thought I'd just post this link here as had trouble finding it the other day :) bit.ly/1XkE9km

btopro:
thank you!

btopro:
#669

btopro:
![icon_header](https://cloud.githubusercontent.com/assets/329735/21322427/f748580a-c5e6-11e6-9f85-541456b3264c.png)
![line_header](https://cloud.githubusercontent.com/assets/329735/21322428/f74bc076-c5e6-11e6-907c-233efddee608.png)

btopro:
inspiration from @rickhumphries 's group

btopro:
![2016-12-19_14-12-30](https://cloud.githubusercontent.com/assets/329735/21325624/3dda2f1a-c5f5-11e6-9ac8-13bdc7a6823c.png)

@rickhumphries can you ask your graphics person to get you an SVG version of this example icon? Also what font is that? I want to start reverse engineering some of these things into possible shortcode templates / etc

rickhumphries:
Will do. The CSS behind that font is: `font-family: 'Sriracha', cursive;`

rickhumphries:
Oh, and the heading is `font-family: 'Special Elite', cursive;`

btopro:
![secondary_source_list](https://cloud.githubusercontent.com/assets/329735/21328634/ee102c1a-c602-11e6-84a6-460eda4341a1.png)
[...]

PR #267, opened 2015-10-01
Foundation Access Template Features File

mmilutinovic1313:
I have added new Image, Video, and Quote properties to the Foundation Access Template Feature File referenced at https://github.com/elmsln/elmsln/issues/206. These changes are made in a new folder in features called foundation_access_template.


btopro:
@heyMP this PR is for the style guide elements currently there. For the sake of showing @mmilutinovic1313 the workflow I told him to name it this but I don't know that its the final name.

I think we should do one of three things w/ the name and packaging of this. These names could be...

- in textbook as that's where other templates are delivered from
- in foundation_access_templates name space since they work best in there
- in some kind of elmsln_style_guide or something project space since they are directly tied to the style guide

so PR looks good if feature is placed in the right project name space which I'll let @heyMP dictate today when you guys meet. Personally I think either foundation access or style guide based is the way to go. FA based would suggest this ONLY works w/ FA, style guide based would imply we could theoretically support new theme engines in the future w/ the same kind of styling (though that's nearly impossible in practice)

heyMP:
I like the namespace `foundation_access_template`.  @mmilutinovic1313 can you change the `package` category of this feature from `Features` to `Foundation Access`

heyMP:
Let also change the all of the images from `../unsplash.jpg` to `https://unsplash.it/1200/800/` so that they render properly.  We can always change this later.

mmilutinovic1313:
@heyMP is the image change done in the WYSIWYG template creation step, or do I have to go in and change the source once the feature is created?

heyMP:
Nice.  Looks good to me!

heyMP:
ugh...  It looks like the WYSIWYG is stripping some elements from the components.  For instance, the `<figure> <figcaption> <aside>` tags are not allowed.  Now that's an easy fix we can add them to the list of allowed tags in the text formats page.

What I don't understand is it's actually completely removing `<a>` tags that are nested within `<figure>` tags.  I'm saving the component and when I go back to edit it the structure is completely different.

heyMP:
Ok I found it.  It looks like we have to include `&nbsp;` in the `<a>` tags that don't have any text in them.  These are used for opening and closing icons for the media component and the imagelightbox.

@mmilutinovic1313 can you include `&nbsp;` in between the `<a>` tags that don't contain text.

For example:

`<a class="mediavideo__open icon-elmsmedia-black" href="#" title="Click to view video."></a>`

- to -

`<a class="mediavideo__open icon-elmsmedia-black" href="#" title="Click to view video.">&nbsp;</a>`

btopro:
@heyMP the default input filter configuration for textbook is currently to remove all unknown tags. So, adding these to textbook's input filter will let them pass through. We may also want to consider dropping the limitations on input as we get closer and closer to HAX. the point of the limits was to enforce accessibility / design (via wyswyg) bc it wouldn't let you do things that were potentially inaccessible.

If we move WYSIYWG into the "advanced" category and that it's no longer the primary method of page production then I think it makes sense to let people do whatever they want in that mode (as we're actively incentivizing NEVER using it while NEVER taking it away)

btopro:
foundation_access_template is singular, just sayin.

also could you do ://unsplash.it/1200/800/ for the links? it's weird I know but I believe browsers will then use the current request protocol in front of it (http for local vagrant stuff and https for live environments)

heyMP:
@btopro I've been hearing that you don't have to do that anymore for https stuff. https://twitter.com/paul_irish/status/588502455530311680

Yeah it might make sense to change it from `foundation_access_template` to `foundation_access_templates`.  It has a much more intuitive connotation.  @mmilutinovic1313 I know it's a pain but would you be able to re-export the feature with the new name?

mmilutinovic1313:
Yep yep. Not a problem. I'm on the case right now!

mmilutinovic1313:
Double Checking before pushing: 

I have added the *&nbsp;* tag to the *image_lightboxed* part of the .inc file. 

I have left the HTTPS on the changed tag after reading the tweet @heyMP posted. -- @btopro and @heyMP Let me know if this is okay, or if you want me to change it

I have renamed the folder in the feature to foundation_access_templates

Let me know if this is all in order, or if I need to change any other things in either the WYSIWYG templates or the exported Features. Additionally, how do I get the code formmating inline for GitHub? I can't seem to figure that out.

btopro:
That sounds correct to me

it's three ` in a row on a line to start it, the code, then three ` in a row on their own line to close it
```
some
code
here
```

heyMP:
@btopro is this pull request good to go?  I'm going to start defining all of the foundation_access view modes in this feature.

btopro:
sorta. you can accept but make sure you delete the duplicate name space; see https://github.com/mmilutinovic1313/elmsln/tree/master/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features

There's currently _template and _templates

Do that and I think we're g2g

btopro:
just realized I can do this and was overthinking. will merge it in then delete the duplicate. @mmilutinovic1313 in the future make sure to delete the old and then issue another commit when there are name space changes like this

btopro:
@heyMP @mmilutinovic1313 good to go. merged then removed the duplicate

Issue #268, opened 2015-10-02
Features Warning Notice: Undefined index: group in views_handler->ui_name()


mmilutinovic1313:
I'm curious if anyone else has run into this Notice message when going to export a feature in ELMS. When I click the *Create Feature* button I get the following error message at the top of the screen: 

    Notice: Undefined index: group in views_handler->ui_name() (line 277 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/views/includes/handlers.inc).

It does not appear to be affecting anything with feature export or the over all User Experience, but I can do a bit of research on why this error may be appearing.

heyMP:
I haven't yet.  I wonder if it has anything to do with the `views_ui` module being disabled by default?

btopro:
agree w/ @heyMP I'm guessing it's related to views UI not being enabled but see if it goes away before and after that happening

btopro:
Can confirm it's happening for me and also happening regardless of Views UI being enabled

mmilutinovic1313:
I can also confirm it happens regardless of enabling.

mmilutinovic1313:
From a cursory glance and a little research, this seems to indicate that there is a value that the code is trying to get that isn't accessable: 

"The cause of this is code trying to access a value (such as 'uid') that hasn't been added to the query."

    $value = $this->get_value($values, 'uid');
    
"get_value() searches for 'uid' among aliases, the aliases array doesn't have such a key, a notice gets thrown."


https://www.drupal.org/node/1103234

mmilutinovic1313:
I have posted in the Views issue queue to try and get this issue resolved. The thread is 5 years old, but was never resolved. I will continue investigating this issue, by monitoring the thread and looking for possible patch solutions.

PR #277, opened 2015-10-06
Fixed Features Warning Notice: Undefined index: group in views_handler->ui_name() #268

mmilutinovic1313:
I have fixed the Features Warning Notice: Undefined index: group in views_handler->ui_name() that was displayed when going to the 'Create New Feature' page. The problem was that the legacy field
   
    field_internal_classification     

was still being used in the cis-7.x-2.x source code. Removing lines that reference the old variable the Warning Notice goes away and the issue is resolved.


btopro:
Fixes #268

PR #285, opened 2015-10-15
Mediavideo and image view modes for external video and image content types

heyMP:
This has been tested locally on a fresh install of ELMS.

## Test
 - Revert the elmsmedia_feature
 - Clear cache on courses and media tools

### Mediavideo
 - Create an external video node with an image
 - Click on the 'share' button and select the token under `Render using template: mediavideo`
 - Embed the token into a page on the `courses` tool
 - Change the `display_mode` from `mediavideo` to `mediavideo__thumbnail`

### Image
- Create an image node
- Click on the 'share' button and select the token under `Render using template: image`
- Embed the token into a page on the `courses` tool
- Change the `display_mode` from `image` to `image__left__lightboxed`

## Unison
I'm also including a folder under `scripts` which contains a setup script for installing a unison profile for ELMSLN.  It's your call if you think it should live here or in the developer repo.


btopro:
"Revert the elmsmedia_feature
Clear cache on courses and media tools"

So the revert can happen either in a update hook or a drush recipe

The clearing all caches in all courses / media tools would get hit as part of the actual upgrade routine, i'm going to make sure the stack specific stuff fires prior to the cron job running in the routine, cron frags caches. Just pushed a commit that'll do that so the cache clearing part shouldn't be needed

Either make an update hook or recipe to do the update. upgrade hook is probably fine though this is exactly what I hate about doing that :) (having a file that's loaded w/ reverts)

btopro:
https://github.com/elmsln/elmsln/blob/master/scripts/upgrade/elmsln-upgrade-sites.sh

heyMP:
Yeah I really don't feel good about putting features revert in update hooks.  It's too destructive.  Maybe we can designate that all features actions will be stored in drush recipes; that will give developers the ability to upgrade modules with `updatedb` without triggering feature reverts.

btopro:
I agree w/ you in most cases; w/ the amount of work we're still doing to media I don't consider it stable. maybe we need to break out the core types / identify things that can't be changed from those that can type of deal. even at that point though we are still currently deciding that there's no hands-off mode to these things

btopro:
this was a poorly worded way of saying you are right

mmilutinovic1313:
@heyMP looking into this to test this PR.

heyMP:
@btopro what is the drush command to create an update recipe?  I can't remember how to do it.

btopro:
drush dwr --for-drup

mmilutinovic1313:
@btopro and @heyMP It works for me! The only thing I found is that if you copy paste from a different font than what is in the WYSIWYG editor (eg. when I copied the text "mediavideo__thumbnail" from this github issue queue) the filters will not render the image. Apart from that looks good thus far!

btopro:
that's somewhat by design; textbook will be removing some of its restrictions soon as the defaults are a bit too strict

heyMP:
@btopro I give up.  I'm not understanding why this drush update recipe isn't working when running the `/scripts/upgrade/elmsln-upgrade-sites.sh` command.

btopro:
two part issue; that script needs modified so that it doesn't have -all cause that's silly. also your script should look something like... https://gist.github.com/btopro/06576bb32d6b9467eb3c

btopro:
so the call is ` drush @${stack}-all drup d7_elmsln_${stack} ${elmsln}/scripts/upgrade/drush_recipes/d7/${stack} --y`

which means drush @media-all drup dr_elmsln_media /var/www/elmsln/scripts/upgrade/drush_recipes/d7/media is the folder it'll look for the upgrade pattern and file pattern in

heyMP:
I added the correct upgrade recipe and tested it against my local.

PR #348, opened 2015-11-17
Applying patch for tickmark ` issue

mmilutinovic1313:
Applying the patch - Hopefully this is the correct way to PR a patch!


btopro:
@mmilutinovic1313 resolve the other issue on d.o. but this works for our purposes / is documented. nice work

Issue #192, opened 2015-08-18
Link to specific course section.


heyMP:
## Problem
When switching between course sections, the section context is stored in a session variable for the user.  They currently have no way of sharing a link to a specific section of the course.  The scenario would be an instructional designer wants a faculty member to review a specific section of their course.

## Solution
We could add a query parameter onto the url that could attempt to switch the current users context to a specific section.
Example: 
`http://courses.elmsln.local/sing100?course_section_context=0001fall2015`

We could also possibly add a share link of some kind to generate the url.  
![screen shot 2015-08-18 at 4 21 15 pm](https://cloud.githubusercontent.com/assets/3428964/9341762/d1a02418-45c5-11e5-817d-87a2ac35a8b0.png)

btopro:
related: #144

btopro:
beyond just a link to the course section I think we'll need (longer term) an interface / screen via @michael-collins that merges the switch section connotation w/ the "section list" that's currently only available to admins. This could be something called upon to swap sections as well as get an overview (staff / instructor side) of how many students are in this section, what's active, etc.

As per #144 and other todo statements in the code we need the ability to effectively masquerade in two distinct ways:
-- Masquerade as a user, across systems
-- Masquerade as a section, across systems

Current proposed solution of something in the url is probably a decent way to go unless we had something like #165 able to track and serve that up (probably url first, then CPR if it makes sense later on).

heyMP:
I'm thinking that this would probably be an appropriate setting to store in a cookie.

btopro:
Have a button that is for "link to this section" / maybe put it in the share menu (which we need to bring back but that's another issue)

`hook_url_inbound_alter` could be used to look for `?elmsln_active_course` in the address and rewrite the address to not have that BUT set the `$SESSION['cis_section_context']` which would be the same as switching the section for someone. (see `_cis_connector_section_context` in cis_connector.module)

To emulate this when switching to another application
`hook_url_outbound_alter` could be used in a similar way to append `?elmsln_active_course` when the hook finds that `$SESSION['cis_section_context']` isset to a value.

btopro:
so this is working, now we just need a share link of some kind

btopro:
@heyMP did you ever implement a place this is shared to?

btopro:
[...]

PR #394, opened 2015-12-09
Link to specific course section. #192

heyMP:


btopro:
blah; travis doesn't seem to like that we make user accounts on demand then execute commands as them during setup

PR #398, opened 2015-12-10
Change method of allowed sections for user.

heyMP:
What if we just used the `cis_section_all_sections()` function to gather the list of allowed sections for this user?  That aligns it more with how the rest of the section switching works.


btopro:
Like the approach though it won't automatically fix someone not being in a section who should be (like staff / admins). We can handle that elsewhere though that on creation of section nodes it needs to add all staff / admins to the group automatically (making a new issue for this now).

PR #403, opened 2015-12-15
Used vmax units to scale the page titles responsively.

btopro:
https://github.com/elmsln/elmsln/compare/responsive-titles?expand=1#diff-ea43b5e2b2a03b889a8a6c79a5efef06R79 seems to be about the only part of this that's actually different between the two


btopro:
app.css gets changed by this which is why this won't merge correctly. @heyMP could you apply the changes from the one file (or at least ensure a really long title works ok now as per this issue). cleaning up some dead branches

btopro:
@heyMP can you verify if this has been accounted for or not? It's SASS so we can't merge directly but the changes to the _content.scss file look minimal

btopro:
closing out, CSS has moved too far past this at this point

PR #440, opened 2016-01-07
Added some more detail to the page

bradallenfisher:
This should probably be broken out into pages of a wiki or something... but this will do for now. Is this heading in the right direction?


bradallenfisher:
Node JS has a nice wiki https://github.com/nodejs/node/wiki

btopro:
@bradallenfisher and we have http://docs.elmsln.org built by read the docs. The scale page(s) are supposed to be in there. You can modify the YAML in https://github.com/elmsln/elmsln/blob/master/mkdocs.yml

btopro:
adding the line for scale to show up now, it should have been there

PR #449, opened 2016-01-08
Dynamic book hompage menu callback

heyMP:
This adds a new menu callback at `mooc/book-homepage` that will look for the current section, and if that section has an outline specified, will redirect the user to that books homepage.


btopro:
`$node = _cis_service_connection_active_outline();` would snag this quickly but needs a refactor (cause obviously it shouldn't have been in there)

PR #457, opened 2016-01-10
Allow empty book outline pages to be the homepage.

heyMP:
Now now that we've added a menu callback `mooc/book-homepage` that dynaimcally switches the homepage depending on what section you are in, we need to allow empty pages redirects on the homepage.


btopro:
ok, I just put that check in Thursday lol

heyMP:
lol... Do you remember why?

btopro:
BC we had a course that put blocks and other content on the homepage. Not a big deal well just add an empty space or something

Wp8

PR #468, opened 2016-01-13
Added Behat Bash Install Script and Supporting Files

mmilutinovic1313:
This commit is the bash script to install Behat - and any supporting files also needed for this process:

composer.json
behat.yml
behat.sh

TODO: Will need to change the FeatureContext.php file to use MinkContext instead of BehatContext manually. This will be updated in a future PR.


btopro:
needs a rebase

Issue #216, opened 2015-08-26
Copy to clipboard button


btopro:
It would be great to have a copy to clipboard button for remixables / other areas where it makes sense to have the user copy and paste something. example from an application I used recently
![2015-08-26_08-48-03](https://cloud.githubusercontent.com/assets/329735/9493659/4a247dd0-4bcf-11e5-944f-ff96bdaab8e1.png)

Assigning heyMP since he might know a codepen that can easily provide this

heyMP:
http://zenorocha.github.io/clipboard.js/

btopro:
Yes, exactly that

zenorocha:
+1 ;)

btopro:
@zenorocha you'll be happy to know this is now in core :) @heyMP should shoot a video once we get the icon @michael-collins made worked in

michael-collins:
@btopro I redesigned the icon and posted to slack (twice). Reads much better, particularly when small in size.

btopro:
saw that, make a new related issue for styling it better so @heyMP doesn't forget. I like the icon btw, also like the way Potter is currently implementing the library (you can hover on an image and click to get the shortcode to generate it in the currently visible style)

PR #511, opened 2016-01-27
Issue 216 copy to clipboard

heyMP:
Implements functionality described in #216 
Waiting on clipboard icon to style.


michael-collins:
@heyMP You should have the icon, via slack. Let me know if it's not right. I designed it with the functionality of copying a paragraph of text to clipboard in mind. If we are copying other things, we may need a more generic design...

Issue #559, opened 2016-02-02
Hide Dont have access to message


btopro:
This shouldn't display for certain roles and seems to throw it up incorrectly at times

btopro:
@heyMP `cis_section_url_inbound_alter` isn't right :(

btopro:
not an infrastructure blocking issue; important but we're very close to bumping up a version

PR #562, opened 2016-02-03
#559

btopro:
@heyMP can you review this please. I noticed that it was possible for people to be given the "YOU ARE NOT IN THIS SECTION" incorrectly. The block is supposed to be preventing me sending you a link that logs you in but really that's like a 0.001% use-case. What we should do is block stuff and then if someone sends someone else a direct link that would switch sections we should add an additional -share prefix or something so we know that's what it is.


heyMP:
This was throwing a white screen when initially testing.  This was because the query parameters and the entity field query were both using the $query variable.  I changed the variable names and it corrected the problem

PR #575, opened 2016-02-04
Update centos7-install.sh use httpd:8080 with varnish

AndrewGearhart:


btopro:
"That file does not exist at that part of the operation" - @btopro

Issue #594, opened 2016-02-08
Document steps after one-line installers


btopro:
If installing against a dev deployment / onto a place like digital ocean, the included Bakery SSO won't work unless the server is aware of the host file change. For example, when you install with
```
yes | yum -y install wget git && git clone https://github.com/elmsln/elmsln.git /var/www/elmsln && bash /var/www/elmsln/scripts/install/handsfree/centos7/centos7-install.sh elmsln ln elmsln.dev http admin@elmsln.dev yes
```

This assumes the server lives at `elmsln.dev` style addresses and so you modify your local /etc/hosts file in order to tell your browser about this. That's great, but the server has no idea of this requirement and I'm not positive if we want to automatically apply this (debating as I just realized this).

At minimum we need to document this fact, I'm going to guess we'll ultimately want to apply it just like vagrant does. If it was installed with real addresses then we won't apply it (so like conditional specifically for this fake address pattern of elmsln.dev).

This is now a blocking issue as far as usability goes from a developer side of things. discovered via conversations with @rickhumphries in #475

btopro:
Commit testing in travis now https://travis-ci.org/elmsln/elmsln/builds/107871027

PR #596, opened 2016-02-08
Automatically modify /etc/hosts on elmsln.dev builds #594

btopro:
The one line installer has elmsln.dev as the default; if this is used it's obviously not for production purposes and so should be treated as a development space (that happens to be out in the cloud). As such, we need the server to know that it's operating this way by automatically modifying it's /etc/hosts file.


btopro:
#594

Issue #577, opened 2016-02-04
Cent7


btopro:
- [x] add vim / nano ootb

btopro:
please feel free to add other things that don't make sense OOTB / might be worth adding in

PR #601, opened 2016-02-10
adding vim & nano to centos7 install

AndrewGearhart:
some sort of text editor should be installed... vim/nano... what camp are you in? should resolve issue #577


btopro:
build failed though a parallel build worked and neither of these influence the actual build lol

Issue #573, opened 2016-02-04
Support for adding external applications to course network


rickhumphries:
It'd be a nice feature to be able to add a link to any external applications that you might have for a course by having an 'add external service' option here:
![image](https://cloud.githubusercontent.com/assets/4973810/12813224/8da92bee-cb2e-11e5-845c-396b8aae0bcd.png)
Although, I know this would be handled in the CIS.

If you had the option of then providing a url and uploading an icon for that service then you can start to bring in external applications that you might want to appear on this menu. For example, we have an separate assessment system and e-portfolio for some courses.

btopro:
poorly named but that's the idea of #407

Currently these can be modified via code pretty easily, though obviously a way of doing it via UI is more empowering.

Examples of some code based changes:
http://stem-researchethics.org/ -- fly out is entirely different from normal as all the courses in their system are OER and related, so we mirror the order.

Overrides in one of our courses to provide deeplinks into parts of Canvas:
![2016-02-04_08-28-59](https://cloud.githubusercontent.com/assets/329735/12816285/78973404-cb19-11e5-8b8b-7aac43b3ea79.png)

rickhumphries:
Ah yeh #407 is exactly what I mean. Can't believe I got a duplicate tag...shameful community behaviour! haha

btopro:
hehe. Ur screenshot / description is far better then 407. I'll close 407 and continue that here.

I've planned this out before but haven't landed on a solution I like but basically you'd have 3 kinds of tools:
- Ones ELMSLN provides on its own / really deep integration
- Ones that you as the admin say you have (like OUR calendar tool, assessment tool, etc) theses can be told to auto be added to all or if people can configure to use them or not
- Ones that people can add on demand (as you suggest) which would be (short term) "my course has a youtube channel so let's add that since I point them there a lot" and longer term is "Let's LTI launch to..." so it becomes easy and complex linkage but is variable

Option 2 would support LTI launching as well (currently we are a provider system that can spider so it's basically faking consuming but will support consume in the future)

There will always be the current method of developer highjacking what's supposed to be there as needed :). Here's an example of some (blarg) logic I use (sanitized). You can see we are pushing to different places based on very specific values right now (which I hate obviously) :)
```
/**
 * Implements hook_cis_lmsless_theme_vars_alter()
 */
function EXAMPLEMODULE_cis_lmsless_theme_vars_alter(&$vars) {
  global $base_url;
  if (isset($vars['services'])) {
[...]

PR #606, opened 2016-02-10
Issue #573

heyMP:


btopro:
Is this a work in progress since I see `tabledrag_example` as a function in one of the items? Just asking before I accept it

Issue #552, opened 2016-02-02
stomp out admin theme mode for non-admin users


btopro:
the administration theme should ONLY apply to paths that are for admins. This is a relatively easy fix in most instances. #551 shows this.

btopro:
- admin/config/development/performance (remove)
- admin/content/book/sar (move / replicate)
- admin/content/book (this will wait til OD 3.x)
- import/feeds_node_helper_book_import (needs moved probably)

PR #621, opened 2016-02-12
#552 dropped admin theme on user/* menus

mmilutinovic1313:
This commit removes drupal from treating user/* from using the admin theme


btopro:
Looks good, next up the js error

Issue #635, opened 2016-02-15
Book nav auto labeling settings not working.


heyMP:
![screen shot 2016-02-15 at 5 20 27 pm](https://cloud.githubusercontent.com/assets/3428964/13061832/7a4e0dc0-d408-11e5-97c5-ce8f6e47652b.png)

heyMP:
PR coming shortly

michael-collins:
I'm pretty sure that the Auto Number should come after Auto Label. For instance. Lesson 1 - Something, Lesson 2 - Another thing. I can't imagine a time where you'd want 1. Lesson - A Thing. 2. Lesson - Another thing. Auto labeling was how we were going to deal with some people wanting to call things lessons and some call them modules, units, etc - yet share the same content.

btopro:
@heyMP yeah I'd think it would be Lesson 2 - whatever.

PR #636, opened 2016-02-15
Book Sibling Nav auto labeling

heyMP:
Issue #635

PR demo:
https://youtu.be/l_0KARCRAZs


btopro:
will accept the PR then switch the order of number and label as @michael-collins suggested

Issue #469, opened 2016-01-13
Update Behat bash Script Settings


mmilutinovic1313:
The behat.sh script needs to add these lines to the FeatureContext.php file: 
   use Behat\MinkExtension\Context\MinkContext;
   MinkContext
I will update the .sh file to insert these lines upon installation.

Additionally - if anyone has other recommendations for extensions for Behat, let me know and I can add them to the composer.json file.

btopro:
@mmilutinovic1313 didn't you do this or is this that thing that optimizes the Behat script?

mmilutinovic1313:
@btopro this is the optimization thing. I'll work on adding this as a patch after I finish the menu bar bug.

PR #665, opened 2016-02-23
Behat Script update Resolved

mmilutinovic1313:
Tweaked behat.sh install script to replace FeaturesContext.php file so that it would use MinkContext


mmilutinovic1313:
https://github.com/elmsln/elmsln/issues/469

Issue #503, opened 2016-01-22
Document shortcodes


btopro:
Bill wrote a great tutorial on the process involved in applying shortcodes / tokenizing a course. wrap that into the docs

btopro:
@IBGregory ask me for the tutorial he wrote so we can get it converted to markdown and cleaned up to match how it works with the current user experience

btopro:
closed from that PR

PR #770, opened 2016-03-21
Document shortcodes #503

proconnor15:
This is the documentation for using tokens.
Issue #503


btopro:
Way too many commits for 1 file but I'll accept this for now. @proconnor15 review https://github.com/elmsln/elmsln/blob/master/mkdocs.yml and see where it would make sense to place the reference to the applying shortcodes documentation

Issue #749, opened 2016-03-15
Play with Features Builder


btopro:
Related to #745 . We've got 745 in the repo and its enabled out of the box but we need to document how to use it and have experience working with it. Basically it allows you to automatically produce features when normally people would have been going through and clicking items one at a time.

btopro:
From talking to @proconnor15 his is a good starting point for #26 but will need san additional layer of automation via task runner to be useful to us outright. Drush plugin it provides is great though

proconnor15:
https://www.drupal.org/node/2699875
Updating so that prefix is necessary.

PR #784, opened 2016-03-28
feature_builder_default as dependency #749

proconnor15:
Added featured_builder_default as a dependency in innovate. #749


btopro:
So the dependency addition is correct but it's in the wrong file. Put it in ulmus_innovate.info

PR #786, opened 2016-03-28
Added the features_builder_default

proconnor15:
#749 Sorry for the extra commits, Mark and I were struggling to revert back after the last pull request.


btopro:
nailed it

PR #806, opened 2016-04-04
Mooc Book Nav Enhancements

heyMP:
Still some work to do but it's a start.  

Note: The icons on the screenshot aren't working on my local.

screen shot 2016-04-04 at 7 21 17 pm


btopro:
dood! so much better!

PR #827, opened 2016-04-13
lor export from blog post as our starting point

proconnor15:
This would be the stuff I think I deleted.


btopro:
.... what did you hit to generate this!? lol

Issue #826, opened 2016-04-12
Add simplify


btopro:
replace jammer with simplify if we can. will have to decouple the strongarm'ed variables.

btopro:
https://www.drupal.org/project/simplify

btopro:
@proconnor15 adding the module into the stack; play with it in a course outline tool by enabling this and disabling jammer. They should do similar things but see if you can clean up node forms and things with it and see how well it works

mmilutinovic1313:
@btopro I see that you added the simplify module to the stack in that commit, do I have to en it via drush? I searched simplify on the module page and nothing came up.

mmilutinovic1313:
Redownloaded it via drush. Let me know if we need to re-add the module into the stack in this issues PR!

mmilutinovic1313:
Cleaning up forms with this module is doable, however certain edit menu items are not able to be hidden:

- Visual Identifiers
- Navigation Block
- Tags
- Hide Content

I'm not sure if we want these items to be hide-able, or if this configuration would be fine. I think Hide Content and even Visual identifiers are important to have, but I am not certain we really need to see Navigation Block or Tags. Let me know what you think and if everything is in order, I will do a PR adding simplify and removing jammer.

mmilutinovic1313:
So with @btopro 's comments on the PR, there are still several things that need to be done
- [ ] Design feature / module that keeps jammer's settings while we implement Simplify
- [ ] Gather Requirements for what we want Simplify to have hidden
- [ ] Decide what to do with system setup for hiding things. (i.e. install and upgrades)

btopro:
recommended steps to accomplish:
- Spin up courses sing100
- Screenshot of the content entry form
- enable simplify / disable jammer at same time
- investigate settings for simplify that clean up interface to similar degree as jammer (screenshot / vid)
- see what module currently has the jammer settings included (I believe it's either foundation_access_ux or elmsln_core)
- propose a PR that converts this feature to use simplify instead of jammer

mmilutinovic1313:
![simplifiedmenu](https://cloud.githubusercontent.com/assets/7243665/15321656/de15a08a-1c05-11e6-8fcb-f66f9ea60cba.png)
![simplifiedmenusettings](https://cloud.githubusercontent.com/assets/7243665/15321657/de20465c-1c05-11e6-92c9-29a31267e659.png)

I believe that these settings would be simplified enough for instructors / content creators to not be overwhelmed with options.

[...]

PR #846, opened 2016-04-27
Removed Jammer #826

proconnor15:
Removed the Jammer module. I implemented the simplify module, where it can hide Authoring information, Text format selection, Publishing options, Revision information, Book outline, and Menu settings.


btopro:
can't accept this as it doesn't have an upgrade path. Need an upgrade path to switch systems, modules, and distributions off of jammer and strongarm'ed jammer settings over to Simplify

Issue #785, opened 2016-03-28
Add more tables / clean up templates available


btopro:
Not exactly this but something like it
http://codepen.io/alimudassar111/pen/idyIq
and
http://codepen.io/ldesanto/pen/pEftw

Also look through the template buttons and make sure we have screenshots for all of them (we don't now) and re-export the associated feature with all the templates in the MOOC distro

mmilutinovic1313:
This sounds good to me. was there a site that we had this updated on last time? I.e. ELMSLN docs. I can't quite remember where we had this stuff stored.

btopro:
@heyMP has a style guide but we aren't including these templates there currently so no just the templates in mooc distro so that the wysiwyg editor has some table options. Also if you know of a better wysiwyg template browser that'd be good ;)

mmilutinovic1313:
This is where my codepen will live. http://codepen.io/mmilutinovic1313/pen/NNYmRy I'm working on it currently while trying to remember how to CSS. I took out the buttons from the second one because I figured editing would be done through the drupal backend itself? If you think we should use the buttons I can readd them and find some images

btopro:
yea the buttons aren't needed;

mmilutinovic1313:
http://codepen.io/mmilutinovic1313/pen/grjxga

New pen here

heyMP:
Nice start.  Can you try another color scheme where the background is white, since the ELMS course content is set on a white background?

mmilutinovic1313:
Updated the background to be white and added a more interesting table color than the previous one!

heyMP:
Lets make it resonsive too.  Can you remove the `width` property or set it to 100%?

Also, can you look into the materialize responsive table plugin? http://materializecss.com/table.html
It looks really simple.  you can use the  materialize javascript library.  https://cdnjs.cloudflare.com/ajax/libs/materialize/0.97.6/js/materialize.min.js

btopro:
getting there

mmilutinovic1313:
@heyMP Would I add that plugin to the JS category on codepen then?

heyMP:
Yeah in the JS settings you can add custom libraries like jQuery.

mmilutinovic1313:
I've made two tables based off the materialize library. Both are responsive. One has stripes, where the other one has a hover feature. Let me know what you guys think of these designs / if you'd like me to design some other ones for MOOC.

mmilutinovic1313:
Whoops! Here are the links: 

[...]

PR #873, opened 2016-05-06
Added a table template #785

mmilutinovic1313:
(exported from the WYSIWYG editor) and two of the sass files for making tables responsive to the foundation_access theme in contrib. I checked to make sure adding these files didn't crash the system, and they don't. If someone can double check that these are in the right place that would be great! This is the first time I've added styling things to the system so I want to make sure I didn't break something :P. Let me know if there is anything i need to change.

https://github.com/elmsln/elmsln/issues/785


mmilutinovic1313:
@heyMP + @btopro

btopro:
ya will need testing after recompiling the sass file, which I can't do ;)

btopro:
what's mooc--table.tpl.php about?

mmilutinovic1313:
That has the table's html code with the imports to make it responsive. Basically I stuck the html from codepen into a template and exported it. I wasn't sure if that was the correct place for that file. But I figured it needed to be put somewhere after export so it could be picked up by the system.

PR #896, opened 2016-05-11
#826

mmilutinovic1313:
- Simplify seemed to be in the modules/ulmus directory so that should be set.
- Removed Jammer from modules/ulmus.


mmilutinovic1313:
Ah I must have to remove jammer from the sh script

btopro:
We'll have to talk through this one. The issue with it wasn't it being there, it's the upgrade path required :). Jammer has settings exported across all the systems currently so it can hide things. We could remove all of those exports and have a module (or elmsln_core feature export improvement) that basically makes it apply those settings so that we don't lose all the jammer settings in the interim.

PR #905, opened 2016-05-17
#826 (mooc_ux_defaults changed for testing)

mmilutinovic1313:
Switched out jammer with simplify for mooc_ux_defaults. Switch didn't brick system and changed settings loaded after re-export. Using this as a start for testing.


btopro:
This looks reasonable; can you switch it from a defaultconfig item to strong arm. Difference is that we can revert (globally) these settings if they are strongarm vs defaultconfig we can't

mmilutinovic1313:
Sounds good. I did a re-export with it as strongarm. I also added menu-links and other permission dependencies. Let me know if it looks good or if I need to remove any of the files.

btopro:
what's the `menu links` or `view hidden fields`?

mmilutinovic1313:
hidefieldsfromuser1

`view hidden fields` is used for the "hide from user 1" option it looks like. 

`menu links` was one of the feature export options that appeared along with simplify. It looked a little bit odd (see picture) as it had a bunch of " ---  " with it. I put it in because I thought we needed to include all of the options for simplify.

btopro:
Oh no drop that I know what that is. Has to do with the menu links that appear in the admin menu (I dono why it's even giving the option to export this). Thanks for clarifying the user 1 thing

mmilutinovic1313:
Should I remove the file too that got generated initially? (menu_links.inc)

btopro:
yes

btopro:
mooc_ux_defaults.features.defaultconfig.inc
 can be dropped too

Issue #836, opened 2016-04-20
increase the number of advagg bundles in elmsln_advagg


btopro:
We right now have 2 of each but really since LMSs have a captive audience we probably should set this higher. Mike Carper recommended 15 or something high. This way it bundles it all up in a way that makes sense but has a higher chance of hitting the same thing

mmilutinovic1313:
Set bundles to 15 because high numbers

mmilutinovic1313:
update hooks: .install file (look in cire fir sexample). Do variable set to equal 15 for each of the names.

btopro:
`elmsln_advagg.install`

mmilutinovic1313:
Should be good to go pending merge

PR #912, opened 2016-05-19
#836

mmilutinovic1313:
Increased the bundler values for css and js from 2 to 15. Also added a .install file with update hooks to update legacy versions.


mmilutinovic1313:
hmmm -- looking into this

mmilutinovic1313:
Ah -- I've been spending too much time in Swift. I need ';' in PHP :P

btopro:
put both of these in `elmsln_advagg_update_7000`

btopro:
it probably would work anyway but can u make the vsets be number instead of string values. `15` instead of `'15'`

btopro:
some internal testing w/ 15 bundles is that it's definitely faster! Is completely against traditional wisdom in this area but whatever :)

Issue #901, opened 2016-05-13
Test php 7


btopro:
It's still a bit early for this but it would be great to get more testing of the system while running php 7. This is mostly getting vagrant up and going, then upgrading php to 7, then start screwing around and see if anything happens that shouldn't.

This patch provides support for php 7:
- https://www.drupal.org/files/issues/drupal-2656548-21-php7.patch
- https://webtatic.com/packages/php70/

I've played with this in vagrant and it works but we need additional testing for sure since this is a major leap in php. it's pretty insanely fast.

btopro:
https://www.drupal.org/node/2656548#comment-11148293

btopro:
/var/www/elmsln/core/cores/drupal-7

mmilutinovic1313:
(y)

mmilutinovic1313:
don't update to php-7 but apply patch

mmilutinovic1313:
Applied the patch and ran a few tests to see if the patch broke anything. I did not notice anything breaking - so I issued a PR.

btopro:
Cool, as this is going to get into drupal core at some point and we won't be implementing PHP 7 support natively for some time I'm going to close this and not accept the PR but let it finish it's travis build to see what travis says. This was the point though so good work :)

PR #913, opened 2016-05-19
#901

mmilutinovic1313:
applied patch to allow for PHP-7 to exist. Did not switch PHP versions. Tested it by flushing caches and building a new course - that it did not brick anything. Also moved patch into the PATCHES dir.


mmilutinovic1313:
Nice! This passed. As @btopro said in the #901 we will not merge this yet. But good to know it didn't crash the install!

Issue #942, opened 2016-06-07
Library_CDN implementation for ELMSLN_CDN Module


mmilutinovic1313:
A few tasks that will have to be taken care of in order to hook up the Library_CDN module with the ELMSLN_CDN module. 

- [x] Download and install Library_CDN & Dependencies.
- [x] Test that the Library & Dependencies do not brick the system.
- [x] Try to use jsdeliver to serve content
- [x] Extend the ELMSLN_CDN module to use the API in Library_CDN to make its local and remote CDN calls.
- [x] Extend the ELMSLN_CDN module to have a checkbox to ENABLE / DISABLE / LOCAL / REMOTE handle its files via the CDN
- [ ] ~~Fix CDNJS & JSDelivr CSS errors~~ #994 (will be fixed and supported in this module)
- [ ]  get Header add Access-Control-Allow-Origin "*" support if needed based on remote jquery making ajax calls
- [ ] write and implement the new easyCDN module to take care of CDN & library loading (#994)
- [ ] Need to update library_info hook because of #996.

#680 
#405

mmilutinovic1313:
https://github.com/jsdelivr/jsdelivr/blob/master/CONTRIBUTING.md

btopro:
it's getting there but boatloads of errors still
- pull latest to your local, I cleaned up a few things
- then got a ton of errors though annyang is rewriting correctly against the non-china CDN location

```
Notice: Undefined index: data in advagg_js_alter() (line 914 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/advagg/advagg.module).
Notice: Undefined index: group in advagg_mod_sort_css_js() (line 1779 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/advagg/advagg_mod/advagg_mod.module).
Notice: Undefined index: group in advagg_mod_sort_css_js() (line 1780 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/advagg/advagg_mod/advagg_mod.module).
Notice: Undefined index: every_page in advagg_mod_sort_css_js() (line 1782 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/advagg/advagg_mod/advagg_mod.module).
[...]

PR #960, opened 2016-06-14
#942 ELMSLN_CDN module update

mmilutinovic1313:
Added JSDelivr Support for the ELMSLN CDN module. Checks to see if library can be found by the `libraries_detect()` function. Then sees if it can match a version found on the CDN. If both match. Then it finds all the files that are on the CDN and sets the URL to be the 'data' property of the javascript array.


mmilutinovic1313:
This is in a place where it can keep being expanded upon. Next step is adding cdnjs support.

btopro:
rdf_mapping should be js_alter

btopro:
not super crazy about requiring registry_autoload and libraries_cdn, will have to review further

mmilutinovic1313:
I could probably just recreate the one or two functions that I used from Libraries CDN and we could eliminate the need for that module entirely

mmilutinovic1313:
I do think that registry autoload does some wear and tear on the performance and speed.

btopro:
ya go for that. I dont want to start treating D7 like D8 til we'd have to (which is the path that starts down)

mmilutinovic1313:
Okay - I'm gonna push the js_alter fix and then I'll work on recreating those functions!

btopro:
if you dropped CDN as a requirement you can't use the constant `CDN_PERM_ADMIN` in hook_menu

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features/elmsln_cdn/elmsln_cdn.module
```diff
@@ -3,3 +3,99 @@
...
+      'file'             => 'elmsln_cdn.admin.inc',
+    );
+
+    return $items;
+  }
+
+/**
+ * Implements hook_module_implements_alter().
+ */
+function elmsln_cdn_module_implements_alter(&$implementations, $hook) {
```
btopro: explain to me why this works and what this is accomplishing? Here's where I get all professor on you since I want to know if you know why this is doing what I said it would ;)

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features/elmsln_cdn/elmsln_cdn.module
```diff
@@ -3,3 +3,99 @@
...
+    $implementations['elmsln_cdn'] = $group;
+  }
+}
+
+/**
+ * Implements hook_js_alter().
+ */
+function elmsln_cdn_js_alter(&$javascript) {
+  foreach ($javascript as $path => $values) {
+    if (strpos(strval($path), 'sites/all/libraries/') !== FALSE) {
```
btopro: probably can be expressed as `strpos(strval($path), 'sites/all/libraries/') === 0` as you are looking to see if its int he first position of the string and not "does it exist somewhere" which the literal might not return.

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features/elmsln_cdn/elmsln_cdn.module
```diff
@@ -3,3 +3,99 @@
...
+        $name = $data['results'][0]['name']; 
+      }
+      $version = $library['version'];
+      if (!empty($library) && !empty($version)) {
+        $cdn_version_arr_elements = getVersions($name);
+        if (!empty($cdn_version_arr_elements)) {
+          foreach ($cdn_version_arr_elements as $element => $value) {
+            if(strpos($version, $cdn_version_arr_elements[$element]['version']) !== FALSE) {
+              $files = $cdn_version_arr_elements[$element]['files'];
+              dpm("Found a File Version Match");
```
btopro: never include `dpm`s in PRs

btopro:
also looks like this still requires libraries_cdn in order to operate is that correct? The name spaces act as includes for class structure which appears to be here so I'm assuming those files are currently required?

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features/elmsln_cdn/elmsln_cdn.module
mmilutinovic1313: From what I had read when I was doing research on the function you told me to add here, it basically says to unset `elmsln_cdn` so that it is removed from the call stack (wherever it may be) and then re-add it again (which adds it to the end of the call stack). This then would make that hook fire last as we want it to.

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features/elmsln_cdn/elmsln_cdn.module
mmilutinovic1313: Noted! Sorry about that. I kept them in so that you might be able to have an easier time looking at it to see if it's working. From what I've seen tho, dpm's basically brick any system w/out devel turned on.

mmilutinovic1313:
I disabled both libraries_cdn and its dependencies and it still worked. What do you mean by "the name spaces act as includes for class structure?"

btopro:
Delete the module from the server and see if it still works

Wp8

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/features/elmsln_cdn/elmsln_cdn.module
btopro: ⭐

mmilutinovic1313:
Deleted it and it looks like it is still working

mmilutinovic1313:
Making the additional changes listed in here and adding support for "NO CDN SELECTED"

btopro:
hmm ok

btopro:
default when enabling the module should be to select the one that supports china fyi. enabling this module is to activate that :). THEN the default setting will be to deliver things over CDN that make sense, if we need to disable on a case by case basis then we can

mmilutinovic1313:
Yep! That makes sense! I think as of now the default is set for the China one (JSDelivr) I tweeted at them to ask them about their process for adding older versions of libraries so we could submit some PR's to them with our current versions and it would work supporting more of our files :)

btopro:
cool. we can also review said libraries and test out using the newer versions. in most cases we are not doing this intentionally by any means

btopro:
ok I see what I was talking about with the name space things. You have 2 files in src/ that you don't need. They have reference to JSDelivr and CDNLocal or something. You never call either of these from what I'm seeing so the objects do nothing / the files aren't needed.

btopro:
ok, just making sure. I'll accept this, delete those two files and tweak / critique from there. This currently isn't enabled by default so it's good to mess with further. Nice work getting it to this point!

btopro:
damnit mark this is in materialize branch, lol

mmilutinovic1313:
Yeah still materialize

mmilutinovic1313:
I have another commit that is ready that fixes those issues.

mmilutinovic1313:
Do you want to meet to merge it into master?

Issue #972, opened 2016-06-21
Update Libraries Module


mmilutinovic1313:
Update Libraries to see if that makes CDN stuff work better.

PR #976, opened 2016-06-22
#972 Libraries Update

mmilutinovic1313:
Updated the Libraries module and tested to make sure it didn't break anything in the materialize branch. Currently we don't have many libraries loaded via the Libraries API as only 3 are detected when looking at this update.


mmilutinovic1313:
FYI - Did this in materialize branch

Issue #982, opened 2016-06-23
Patch FastClick module to support versioning


mmilutinovic1313:
The FastClick module should be patched to support actual versioning. Currently it has its version listed as master, which is a useless parameter as no CDN or external module can use that non-numeric version to interpret what version to use. This is pretty low priority, but will be very easy to patch.

mmilutinovic1313:
Patched the module at https://www.drupal.org/node/2755133. Added patch and updated our codebase. After checks and merge, this issue should be good to be closed

PR #985, opened 2016-06-24
#982 Version Regex Patch

mmilutinovic1313:
Added version regex support @ https://www.drupal.org/node/2755133.


mmilutinovic1313:
Do I need to write an update hook?

mmilutinovic1313:
Ah...JSDelivr fixed got pushed into here too. Meant to put that in a different PR

Issue #991, opened 2016-06-27
Figure Out Why Angular won't load in CDN Module


mmilutinovic1313:
This is related to #942 and #405. We have angular loading locally properly but for whatever reason the CDN is trying to load .js.map files for angular which is preventing angular from loading at all. Got to find the reason behind this. Attaching screenshot for reference.

mmilutinovic1313:
I'm going to double check that this is still occurring. This may have been fixed in #996

mmilutinovic1313:
@heyMP what was the command that you used to fake the system that the angular library was readable by Libraries API? When I blew away my last vagrant, the link went away

heyMP:
I don't remember the exact command.  You need to create a symlink from the angular library into your stack's libraries directory.  So it would be something like:

`ln -s ~/elmsln/core/dslmcode/shared/drupal-7.x/libraries/angularjs ~/elmsln/domains/online/sites/all/libraries/angularjs

btopro:
Oh come over tomorrow and we can run the developer command to merge this in w. Have the script run it

mmilutinovic1313:
Sounds good - I'm currently in and I can come over whenever you get in!

btopro:
this got fixed by the previous commit

Issue #978, opened 2016-06-22
Angular Local Integration


mmilutinovic1313:
So I've been playing with a few modules that say that they support some form of Angular integration. The ones I've looked at all load angular via a CDN however (#977), which is not necessarily what we want to happen in case a CDN is / gets blocked in a certain region (i.e. China). 

My big question is how could I structure a folder in our libraries directory on drupal -- in order to be readable by the Libraries API? Would I need to create a module to do this? Could I do it manually by adding certain files?

btopro:
`core/dslmcode/drupal-7.x/libraries/angular` or something. module to implement it yes. hook_libraries_api or whatever it is

mmilutinovic1313:
I've set up the library in `core/dslmcode/drupal-7.x/libraries/angular` and Libraries API can detect it, but Drupal is yelling at me that I don't have it included in the `@online` specific libraries folder. Do I need to symlink it or something so it gets sent to all the systems to be fully readable?

mmilutinovic1313:
Double checking that the libraries are loading correctly locally. Will make a PR after these checks go through. @btopro: you may have to run the elmsln-sync-lib command after the PR goes through to make sure the library gets picked up properly. 

Eventually we want to write an API for loading the angular libraries instead of just loading it in the cis_dashboard module.

mmilutinovic1313:
This is fixed in #996

Issue #405, opened 2015-12-16
CIS 2.x UX overhaul


btopro:
The UX of the CIS needs overhauled. @michael-collins has mocked things up that we haven't had time to implement to this point. Time to push this up in priority a bit or at least increase visibility of the issue. The current long form that you are dumped to has too many options and is ugly.

http://codepen.io/michael-collins/pen/JYWBvP

btopro:
0.0.2 includes a lot of clean up to CIS based on feedback from IDs. Menus and options cleaned up and all types set in alignment with 2.x backend system style

btopro:
#669 #624

btopro:
feedback from @kat-wehr 
- Network pane in CIS 2.x should have quick CTools modal style pop ups to do this administration
- Course Outline - Be able to drive into the outlines right from there with quick links to outline designer display
- Studio - Assignments overview, adding more, etc
- Discuss - Add more discussions, etc
- Blog - Add more postings, etc
- Innovate - Just go to (basically)


Then per section you'll be able to associate them as to which they use. CIS starts to become the window into the universes while still allowing you to be in those universes to get your tasks done

btopro:
- [ ] abstract current implementation in foundation_access so it's only loaded on the dashboard in cis
- [ ] use angular over CDN instead of local copy
- [ ] roll into master but don't link to

This will allow us to progressively enhance the dashboard or even make a dashboard / management link with a (beta) flag or something next to it. Once the sorting and filtering capabilities are activated it'll be pretty amazing just from playing with it in the current state

mmilutinovic1313:
Quick question about Angular. Are we using Angular 2? Or just the original?

btopro:
1.x for now

mmilutinovic1313:
@heyMP got a bunch of stuff fixed for this yesterday. It's now a bit closer to where master is so that the theme isn't screwed up like I was describing. I'm working now on getting some visuals added and implementing the design that @michael-collins drew up.

mmilutinovic1313:
Adding the checklist here so I can check items off: 

- [ ] abstract current implementation in foundation_access so it's only loaded on the dashboard in cis 
- [ ] use angular over CDN instead of local copy
- [ ] roll into master but don't link to

Currently the materialize branch mirrors @michael-collins ' markup for the dashboard. I have a few questions about the direction in which I need to go to finish off some of these remaining tasks

[...]

PR #992, opened 2016-06-28
Added local Angular Libraries Support #405 & #978

mmilutinovic1313:
Added local support for Angular libraries. We will be using version 1.5.7 for now to start. This is now a library that is detectable by the Libraries API module and will eventually be able to be used with our CDN module (#991). We will need to run the `drush elms-sync-libs` command in order to hook the library into all the systems. This has the full angular library - so if we ever want to cut down on size, we can figure out which files we need / want to load.


btopro:
uhhh yeah are you just trying to improve your line additions? this is 100k lines, what all is it @heyMP and why is it this big? Don't we just need the compressed versions and not the entire repo?

mmilutinovic1313:
@btopro -- yeah the 100k lines is because angular (the library itself in its entirety) is a very big library. We could trim it down if you would like to just small essential files. Also with the easyCDN module I'm working on this should help prevent oversized libraries from always needing to be localized in their entireties.

heyMP:
Yeah let's at least get rid of the angular/docs.  That takes up most of the file additions.

mmilutinovic1313:
Removed that folder so that will reduce some of the lines of code / amount of space it takes up

btopro:
still claims to be 102k additions. redo the commit again so that addition of a billion files isnt forever in peoples histories lol

mmilutinovic1313:
Okay sounds good - I'm gonna close this PR and just redo them

PR #995, opened 2016-06-30
#405 removed angular.js file in foundation access

mmilutinovic1313:
This closes out the #405 issue in the materialize branch. Next step will be merging materialize back into master


mmilutinovic1313:
The only two files that the module that used angular originally loader were angular.js and angular-resource.js. So we could honestly just use those two files (minified & regular) and the versionfile -- and then while I work on the easyCDN module we will have more options in downloading other required angularjs files.

PR #996, opened 2016-06-30
#978 Added Angular Locally (Removed doc Folder)

mmilutinovic1313:
Added Angular properly without the oversized documentation folder. This will still be a large commit, however this saves 50,000 lines of code.


mmilutinovic1313:
Okay...well this is still huge...I'm gonna talk to @heyMP and see if we can just toss more files]

mmilutinovic1313:
Also removed the locale files dir. Those files took up a lot of room.

mmilutinovic1313:
The only two files that the module that used angular originally loader were angular.js and angular-resource.js. So we could honestly just use those two files (minified & regular) and the versionfile -- and then while I work on the easyCDN module we will have more options in downloading other required angularjs files.

Issue #931, opened 2016-06-02
minor bugs to track down


btopro:
viewing a node in mooc after applying simplify (dono if that's the cause tho) node/169
Message	Notice: Undefined index: format in wysiwyg_pre_render_text_format() (line 222 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/wysiwyg/wysiwyg.module).

`admin/structure/features/mooc_ux_defaults` path in mooc
Message	Warning: Invalid argument supplied for foreach() in features_override_module_component_overrides() (line 230 of /var/www/elmsln/core/dslmcode/shared/drupal-7.x/modules/ulmus/features_override/features_override.export.inc).

very low priority but worth noting when we get in squashing mode

mmilutinovic1313:
Gonna check this out today as I'm in bug squashing mode

mmilutinovic1313:
Hmmm I'm not seeing this error when I try and view a node. Checked to see that simplify was enabled. Any other modules that should be enabled or disabled you think?

btopro:
neither one? there's 2

mmilutinovic1313:
Yeah for whatever reason — when I got to the path `http://courses.elmsln.local/sing100/node/(nid)` I don’t get either of those errors

btopro:
and simplify is applied?

Also the other path is `admin/structure/features/mooc_ux_defaults`

mmilutinovic1313:
Yep yep -- I am going to clear caches and such to make sure. Maybe I can mess with the settings and in simplify and see if that's why you saw something?

mmilutinovic1313:
Ah I got the second one after clearing caches and going into the `Review Overrides` section. Still can't seem to find the issue with `/node/(nid)`

mmilutinovic1313:
Should this line:
 `if (!isset($normal_copy[$name])) {`
be:
 'if (!isset($default_copy[$name])) {'
instead?

mmilutinovic1313:
Oh I see! It's if the normal_copy is not set for the $name (i.e. it doesn't exist in that prior foreach) we add it.

btopro:
Sounds like a Good thing :)

Maybe submit that patch to the features_override module if it resolves that 1 notice on cache clear

Wp8
________________________________

mmilutinovic1313:
I'll add this to their issue queue then! Shall I put the patch in our folder for the module as well?

btopro:
nice work @mmilutinovic1313

PR #997, opened 2016-07-01
#931 Second Bug in Issue Sqashed

mmilutinovic1313:
Added the check for `$default` being a valid datatype to be put through the foreach loop. Also closed the first foreach loop for validation. Still need to see if the First bug exists for the issue (#931)


btopro:
see comment on commit

mmilutinovic1313:
Added the changes you had recommended as well as the patch file I submitted to d.o.

PR #1003, opened 2016-07-06
Added an install script for nodejs.

heyMP:
Needs testing.  Should install Node Version Manager, then the latest version of NPM.

## To Test

run `bash elmsln/scripts/frontend/frontend-install.sh`
then `npm`

You should see npm documentation in the terminal.


btopro:
@heyMP can you write a doc page about how / why to use this in the developer section

Issue #999, opened 2016-07-03
review hook_set_cis_service_data


btopro:
see if it's still needed given the snake paradigm.

mmilutinovic1313:
Found the function and am removing it to test. I'm planning on creating a course to test. Are there other specific tests you would like done to test that this does not break?

btopro:
need to look around for things that implement the hook. I know it happens some places like in the mooc distro but that might be the only one

mmilutinovic1313:
Okay sounds good -- I'll search for its implementation then and remove those as well before testing.

mmilutinovic1313:
Tested to see if anything broke and removed the calls to this hook in the `cis_service_connection.module` file. Nothing broke so it looks like we are good to go. I'm making a PR now for this. If you'd like me to run any other specific tests before accepting the PR, let me know!

btopro:
did u find where the hook is implemented elsewhere in the code? It has to exist somewhere

mmilutinovic1313:
Did a search both on all my local files and with what is on github. Only place I can find that hook being referenced is at `core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_service_connection/cis_service_connection.module`

PR #1016, opened 2016-07-11
#999 hook_set_cis_service_data no longer needed

mmilutinovic1313:
Removed the hook from the `api.php` file  and any calls to it from the `.module` file. No breaks to the system found. Only location that I can find `set_cis_service_data` being found in the system is at `core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_service_connection/cis_service_connection.module`


btopro:
excellent work, I was able to track down that it is used in some of our internal implementation code but that code doesn't do anything useful (anymore) :) good work

Issue #987, opened 2016-06-24
Ubuntu 14 Install doesn't seem to work.


vrracing:
Fired up a virgin Ubuntu 14.04 instance at DigitalOcean, a 512mb instance.

Copied and ran the [U1404 instructions](https://github.com/elmsln/elmsln/blob/master/scripts/install/handsfree/ubuntu14/example-deploy) as root.

Note that line 2 says there are 3 lines to be run but there appear to be only 2.

Enhanced the host file with the outputted lines as instructed.

If I go to http://lor.elmsln.dev/   I get what appears to be a drupal installation page with this at the top...

![image](https://cloud.githubusercontent.com/assets/2730237/16353238/1b2c7ce6-3a3d-11e6-9c02-765c17bad6ab.png)

Any suggestions?

I apologize if this is the wrong place but I could see no links to a Google Group or mailing list for the project only "Community/Discussions" which brought me here.

Thanks,
Jim

btopro:
Need 2 gigs to install and go to online.elmsln.dev after setup. Right place to ask :)

Wp8
________________________________

vrracing:
Ah cool.  Thx.  I'd watched one of the YouTubes and noticed the minimal config was selected so just followed along.  Thx, I'll give it another go tomorrow.

vrracing:
RAM isn't the issue.  I tried it again with a new DigitalOcean droplet with 2 GB Memory / 40 GB Disk / NYC2 - Ubuntu 14.04.4 x64 but got the same results.  I followed the directions and copy and pasted the two commands into the terminal as root with no edits.  Here are some error highlights in case there are clues...

No obvious errors until MySQL install then...

```
You can test the MySQL daemon with mysql-test-run.pl
cd /usr/mysql-test ; perl mysql-test-run.pl

Please report any problems at http://bugs.mysql.com/

ERROR 1396 (HY000) at line 1: Operation DROP USER failed for ''@'localhost'
ERROR 1396 (HY000) at line 1: Operation DROP USER failed for ''@'elmsln1'
ERROR 1008 (HY000) at line 1: Can't drop database 'test'; database doesn't exist
Cloning into 'config'...
```

Not a MySQL admin by night so don't know if these are critical.

Next git seems to go and there are some warnings...

```
Checking connectivity... done.
Initialized empty Git repository in /var/www/elmsln/config/.git/

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

[...]

PR #1035, opened 2016-07-15
Root composer drush fix

bradallenfisher:
This pull request is related to #987 
There are still some errors related to:

```
ln: failed to create symbolic link ‘/usr/share/drush’: File exists
unlink(/root/.config/composer/vendor/drush/drush/lib/package.xml): No[warning]
such file or directory drush.inc:691
'drush' cache was cleared.                                                                                                                                        [success]
groupadd: group 'elmsln' already exists
cp: cannot stat ‘/root/.composer’: No such file or directory
```
where now composer, when installed by root is in /root/.config/composer

and

```
/var/www/elmsln/scripts/install/users/elmsln-admin-user.sh: line 61: drush: command not found
/var/www/elmsln/scripts/install/users/elmsln-admin-user.sh: line 63: drush: command not found
```

where drush isn't installed yet at this point.

while these errors are present they do not affect elmsln installation. We can open another issue to clean up these errors if necessary. 

Since drush installing authority tool: blah is now working i say you can pull this in.


btopro:
this fixes / accounts for #987

Issue #1036, opened 2016-07-18
review cachability of all views


btopro:
easy performance gain having this set correctly.

Mark this is higher priority then other past tasks. Find views in different distros and review their usage of caching in views. If they aren't being cached but seem like they could / should be (some data very static, other data incredibly dynamic; most static tho) then start busting out the PRs for feature revisions / update hooks to make it happen

mmilutinovic1313:
Will start churning out the PRs today. Will Devel give me the caching information I need to make the determination if its being cached?

btopro:
if you look in views it has cache output or cache query and lets you select how long for each (don't do authcache at this time if that setting is available). Open a view, hit advanced, it's near the bottom right corner of that list of things. Then figure out what's a good interval to cache (it has like an hour, 6 hours, a day, 6 days). Most likely, 6 days is good for both. Question to ask as to when it's ok to cache that long:
- Did students create it / will need to engage with it? (cache low, like an hour)
- Did staff create it / is it like content in the course itself (a video, a timeline, etc)? (cache high)
- Is it a display people will expect to be updated after they do something (media homepage for example) don't cache or cache very low (like an hour)

mmilutinovic1313:
Studio Choices:

`Assignment bulk edit` = 6 days
`Assignments` = 6 hours
`CIS Section list` = 6 days
`CIS User list` = 6 days
`CLE my submissions` = 30 min
`CLE submission` =  30 min
`Critique - critiques received` = 30 min 
`Critique EVAs` = 30 min 
`Critique feedback given` = 30 min
`flag_bookmarks` = 1 hour
`My Submissions` = 30 min
`OG all user group content` = 1 hour
`OG content` = 1 hour
`OG members` = 6 days
`OG members admin` = 6 days
`OG User groups` = 6 days
`Submissions` = 5 min
`Submissions listing` = 5 min
`flag_bookmarks_tab` = 6 days

(There are also 1 min, 5 min, 30 min options)

mmilutinovic1313:
Innovate Choices: 

`CIS Section list` = 6 days
`CIS User list` =  6 days
`OG all user group content` = 1 hour
`OG content` = 1 hour 
`OG members` = 6 days
`OG members admin` = 6 days
`OG User groups` = 6 days

mmilutinovic1313:
Discuss Choices: 

`Administration: Posts` = 1 hour
`Administration: Threads` = 1 hour
`Category sub-category listing`= 1 hour
[...]

PR #1062, opened 2016-07-22
#1036 Discuss Caching

mmilutinovic1313:
Adding Caching to Discuss. Added Update Hooks.


core/dslmcode/stacks/discuss/sites/all/modules/local_contrib/harmony_core/harmony_core.install
```diff
@@ -634,3 +634,10 @@ function harmony_core_update_7103() {
     variable_set('harmony_core_rebuild_actions', $actions);
   }
 }
+
+/**
+ * Revert cis_displays component in views. This is for caching update. 
+ */
+function harmony_core_update_7104(&$sandbox) {
```
btopro: well need to do soemthing special to modify this view since its in harmony core which is actually out in contrib and not something we have direct ownership of

PR #1100, opened 2016-08-06
0.4.x

btopro:


btopro:
bringing other branches together into a 0.4.x since that's the branch we're working against to later pull into master

PR #1144, opened 2016-08-12
Spot check on pr

### LINKED ISSUE #736: Update Vagrant to use Rhel 7 Based OS and php-fpm
I noticed while working on discuss that the vagrant version uses mod_php. Since elms is working on rhel7, I think we should move everything in infrastructure towards the latest if possible.

bradallenfisher:
Can you accept this so I can do a quick spot check? I had to test this against my own repo since there were some absolute urls in the vagrant up command.

Should fix #736


btopro:
there ya go; we'll want to transition the box from your name space to the elmsln name space if this works

PR #1199, opened 2016-08-22
Added example of Forking workflow

heyMP:


btopro:
g2g; try to work against 0.5.x in the fuure

PR #1218, opened 2016-09-07
fixes centos7 drush and group issues

bradallenfisher:
basically the proper elevated permissions weren't being set after the rhel7/centos7 update occurred. While this isn't the perfect or most elegant solution it fixes the problem without disrupting anything system wide.

### DIFF
--- a/scripts/vagrant/handsfree-vagrant.sh
+++ b/scripts/vagrant/handsfree-vagrant.sh
@@ -60,7 +60,9 @@ drush @online en vagrant_cis_dev cis_example_cis --y
 # restart apache to frag some caches because of the different settings that changed inside
 service httpd restart
 
-su - vagrant 'bash -s' /var/www/elmsln/scripts/install/users/elmsln-admin-user.sh /home/vagrant
+su - vagrant bash /var/www/elmsln/scripts/install/users/elmsln-admin-user.sh /home/vagrant
+
+sed -i '1i export PATH="/home/vagrant/.config/composer/vendor/bin:$PATH"' /home/vagrant/.bashrc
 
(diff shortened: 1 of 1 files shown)


bradallenfisher:
I am not sure why all the other commits are in this merge request. Sorry about that. I just wanted to get the hands-free stuff in.

scripts/vagrant/handsfree-vagrant.sh
```diff
@@ -58,4 +58,26 @@ drush @online dis seckit --y
 # specific stuff for aiding in development
 drush @online en vagrant_cis_dev cis_example_cis --y
 # restart apache to frag some caches because of the different settings that changed inside
-sudo /etc/init.d/httpd restart
+service httpd restart
+
+su - vagrant bash /var/www/elmsln/scripts/install/users/elmsln-admin-user.sh /home/vagrant
+
+sed -i '1i export PATH="/home/vagrant/.config/composer/vendor/bin:$PATH"' /home/vagrant/.bashrc
+
```
btopro: Does this need to happen via su as well?

btopro: The sed and PATH part I am referencing herr

btopro:
- I rebased 0.5.x against master since it had some minor fixes (which you then saw as commits)
- then rebased your PR (didn't know that was a thing but you can click edit and do it)
- then I pushed these commits against master

This is different from the workflow previously proposed only because
- this fixes issues with the thing we claim is the best working configuration (master)
- this fixes things that directly relate to those developing the platform (so best to have it both places).

I rebased this fix into 0.5.x so it's in either branch

Issue #1172, opened 2016-08-17
Review pathcache


btopro:
https://www.drupal.org/project/pathcache

btopro:
Test plan:

- Make a site that has 300 nodes in an outline designer content outline (single outline, can be nested however)
- Enable Devel with the query / page timer / memory usage output appended to bottom of page
- blow caches away, see counts, then refresh and see counts. Go go next page in outline and see counts on initial vs next load.
- enable pathcache and repeat the previous step to see what the differences are if there are any

I've had a few people say this makes a difference but I'd like to know what difference and how significant. This could be something worth enabling on all systems but would like to get some data to support that move / upgrade path (cause we'd have to inject something into shared_settings.php in order to do it)

mmilutinovic1313:
I'll work on this alongside the `hidden_nodes` patch tomorrow. I did find that devel started crashing my machine's vagrant when I was using more and more of its settings. I got a 503 error at one point.

btopro:
@bradallenfisher would be the one to talk to about that but it might also be ur settings. only need memory / page time / query log

mmilutinovic1313:
Will do -

bradallenfisher:
The general rule with devel is turn it off when you aren't using it. So basically don't turn on the profiling stuff when you are doing things you aren't profiling. 

For instance, if you are spinning up courses and stuff, disable memory / page time / query log, until you actually want to look at that data.

mmilutinovic1313:
Ah okay - that makes sense. It shouldn't really crash for this test then since the tests I was running last specifically focuses on caching when the content was updated / created. Thanks! I'll get working on this then. I'll probably use devel_generate to make those 300 nodes.

mmilutinovic1313:
# Stats:

## Without Pathcache enabled

Initial Cache Clear:

> 1108 queries in 588.9 ms | Page execution time was 1912.97 ms. | Memory used at devel_boot()=0.83 MB, devel_shutdown =47.96, PHP peak=51.25 MB

Initial Refresh: 

> 246 queries in 105.74 ms | Page execution time was 227.34 ms. | Memory used at devel_boot()=1.01 MB, devel_shutdown =11.93, PHP peak=13 MB

Next Page Initial: 

> 391 queries in 400.88 ms | Page execution time was 647.36 ms. | Memory used at devel_boot()=1.01 MB, devel_shutdown =13.75, PHP peak=14.75 MB

Next Page Refresh: 

[...]

PR #1257, opened 2016-09-16
#1172 Update Hook & Added Pathcache Settings in shared_settings.php

mmilutinovic1313:
Update hook for elmsln_core.install and adjusted shared_settings.php to include the proper file path for pathcache


mmilutinovic1313:
Should just be one commit...it's doing that weird thing again. Only 2 file changes so should be good

Issue #1015, opened 2016-07-11
My Journal Labeling in Blogs


mmilutinovic1313:
**Version**
0.3.1

**Steps to reproduce the bug**
1. Build a New Course
2. Navigate to the blogs.elmsln.local
3. Look at the main menu items.

**Expected behaviour**
"My Journal" and "writing-exercises" should only appear in nav bar because as their own links as different labels .

**Actual behaviour**
"My Journal" is displaying twice for both "My Journal" and "Writing Exercises". This causes confusion when reading the menu bar. Additionally, the my journal URL is being called within itself in the breadcrumb menu. I am unsure if either of these are intentional.

**Additional information**
OS: OSX El Capitan 10.11.3
Browser: Chrome

This bug should be low priority.

mmilutinovic1313:
![writing_exercises](https://cloud.githubusercontent.com/assets/7243665/16738874/8af9ae24-4765-11e6-8781-63e0ac6c6dcf.png)
![confusing_journal](https://cloud.githubusercontent.com/assets/7243665/16738876/8f70a354-4765-11e6-9d07-b6ca53ec1510.png)

btopro:
has to do with the menu structure most likely

mmilutinovic1313:
So I think this makes more sense being labeled "My Writing Exercises" as the path states that this link should be "writing-exercises". I've changed the name in the menu and view to match this.

mmilutinovic1313:
Should be good to go after merge

PR #1259, opened 2016-09-19
#1015 Fixed Repetitive 'Journal' Label 

mmilutinovic1313:
Journal was listed twice in views and in the menu items. Now it reads  'My Writing Exercises' like the url link states it should be.


mmilutinovic1313:
I guess squash cuz it's adding old dumb commits again...

PR #1260, opened 2016-09-19
fixes #849 replaces internal node links when copying book

### LINKED ISSUE #849: On duplicate check for internal links and rewrite them
When replicating an outline we need to look for the old links and update them to potentially new ones when they reference internal `node/123` thing

bradallenfisher:
@heyMP helped A LOT with this. But it's on my machine so I am pushing it up.

### DIFF
--- a/config/.gitignore
+++ b/config/.gitignore
@@ -2,3 +2,15 @@ dwsync.xml
 _notes
 _notes/
 .DS_Store
+stacks
+stacks/*
+private_files
+private_files/*
+jobs
+jobs/*
+_nondrupal/piwik
+_nondrupal/piwik/*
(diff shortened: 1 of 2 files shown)


btopro:
1. Does this work? Looks like it probably does.
2. It looks possible that old item node/123 would replace references for node/1234 since node/123 will still be found in that. The preg should ensure that it's more of a node/123" type pattern
3. This will timeout on large jobs since it's not doing batch processing. Make the 2nd step work more like the 1st as far as only processing like 10 at a time. You can build the map out first just like you are, but then we need to make sure we're only node_load / node_save like 10 items at a time. This avoids timeouts on very large amounts of nodes (where we'd run into these potentially issues in the first place)

Issue #1256, opened 2016-09-16
Implement views_content_cache for CIS


mmilutinovic1313:
We need to implement `views_content_cache` in CIS to set the caching for these views to be content based and to be set at the highest time increment (6 days). This will be applied to all view types that are visible to users. For example -- Field Collections will not be cached in this way as they will be hit / cached when their content is updated.

btopro:
changed title; this is to fix views caching across all systems

mmilutinovic1313:
Awesome -- After adding the caching, will I be exporting the view like I did the last time we played with caching options?

btopro:
yup, same thing, basically switching them all over to this plugin cause this is soooo much more logical

mmilutinovic1313:
Okay commenting the views from each distro below as a check off list for this task:

mmilutinovic1313:
# MOOC / Course

- [ ] Book Search and Replace (Content)
- [ ] CIS Section list (Content)
- [ ] CIS User list (User)
- [ ] Feeds log (Feeds log)
- [ ] Feeds Node Helper Export (Content)
- [ ] MOOC book roots (Content)
- [ ] OG User groups (OG membership)

mmilutinovic1313:
@btopro: for courses, there is a OG caching option. It has two options of _Current Group_ and _Member's Group_. Would you like me to use either of these to cache the OG views? There aren't too many content-based cacheable views within MOOC/Courses it seems.

PR #1261, opened 2016-09-19
#1256 CIS / Online Views Updated to Content-Caching

mmilutinovic1313:
Content Caching now standard on CIS distro


mmilutinovic1313:
Screenshotted problems with duplicate commits popping up on PRs -- will submit to github later today or this week.

btopro:
Did you change the caching to any node type change or just the types that have influence of that view?

mmilutinovic1313:
I did any node type change. I can do it with influence though if that is the preferred method?

btopro:
they should only flush IF nodes that show up there are refreshed. For example, if a new unit is added, it shouldn't kill the offerings view. If a new course is added / updated, then ya

mmilutinovic1313:
Hmm okay then -- that makes sense. I'll readjust those then to be set-up this way

btopro:
closing without merge, we talked about this previously

mmilutinovic1313:
Yup sounds good to me --

PR #1279, opened 2016-09-29
Materializeoverhaul

btopro:
wow. I can't believe this merges cleanly


btopro:
materialized all the things is now merged into 0.5.x branch with a surprise addition. This is all iterative design so as always, challenge these conventions, reflect usage patterns, and try to offer critique for enhancement in later versions. This is looking pretty solid though there's some clean up to do I'm sure

PR #1290, opened 2016-10-04
0.5.x

bradallenfisher:
We will need to handle updates as well. But it will need to possibly account for ubuntu? Not sure though since I haven't run into it on that flavor. This at least fixes the 503's for new installs.


btopro:
nice

PR #1324, opened 2016-10-11
Changing the path to include new composer

bradallenfisher:
This works now


btopro:
this is way old dont need to support fixing this

PR #1422, opened 2016-10-28
UBUNTU 16 PHP 7 TUNED ETC...

bradallenfisher:
I spun up centos7 and it still worked with all the crontab changes. 

I will test ubuntu 14 as well, but it shouldn't bork. ;)


btopro:
well this is insane that I'm about to accept Ubuntu 16 with PHP7, MariaDB, Apache2.4 w/ HTTP2 support into 0.5.x when I didn't expect it till around 0.8.x.

Obviously more testing to do but wow. incredibly promising. You need to record a video / Drupal planet post or something cause this is nuts!

PR #1435, opened 2016-11-02
Installation Order - swap epel and remi

rickhumphries:
When I've tried installing, can't get the epel repo until you get the remi one. So swapped them in the installation order

### DIFF
--- a/scripts/install/handsfree/centos/example-deploy
+++ b/scripts/install/handsfree/centos/example-deploy
@@ -10,10 +10,10 @@
 # the real deal to be deployed somewhere.
 # get the epel and remi repo listings so we can get additional packages like mcrypt
 yes | yum -y install wget git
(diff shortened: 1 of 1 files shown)


btopro:
Interesting, so this wasn't working in cent 6 installer? Does the one liner fail without this? We moved to cent 7 for vagrant development and cent 6 is pretty old at this point so haven't tried in awhile.

rickhumphries:
I'm not sure if the actual one liner fails to be honest. I was going through each command at a time and just noticed that you can't run `rpm -Uvh remi-release-6*.rpm` until you've run `rpm -Uvh epel-release-6*.rpm` first, but I sorted it before I got to the one liner. Thought it was just a (very) minor barrier to entry

btopro:
ohhh i see, yeah cause there's some silliness w/ a true 1 liner and cent6 bc of something similar to what you are describing.

Issue #1234, opened 2016-09-12
Make our own CDN


btopro:
This could work with #1203 so that we could choose a mode for deployment that forcibly rewrites ALL drupal / elmsln delivered style sheets to come from a CDN such as raw-git / etc. This would mean disabling advagg and instead delivering something from our own CDN. Till then / as a prototype of what would be required, we could hit point release based rawgit addresses such as: https://cdn.rawgit.com/elmsln/elmsln/0.4.0/core/dslmcode/cores/drupal-7/modules/book/book.css

This would mean we'd be rewriting (via EasyCDN) references to the Drupal core book.css and instead point to the rawgit cdn. We'd have to disable advagg as mentioned, so there would be ~25-50 CSS/JS files being loaded BUT they'd only be downloaded once. Then every site you go to anywhere / everywhere (across deployments) would not need to download new files _so long as they are on the same version_.

This would also require either setting the version it's on or allowing for "developmental" addresses as well (in interim that would be https://rawgit.com/elmsln/elmsln/0.4.0/core/dslmcode/cores/drupal-7/modules/book/book.css ; not the cdn. is missing which prevents cloudflare hit)

btopro:
Might be a great use for the aws pilot. Cloud based delivery of files that have no Pii implications

btopro:
#994

bradallenfisher:
just to be sure we are on the same page here...
Do we really want to build a CDN? Or would using a service be a better option?
AWS Cloudfront seems promising. But before I look into it, I want to make sure there aren't other reasons we would want to build our own.

btopro:
Not going to be doing this, the new webcomponent paradigm makes asset delivery make so much sense that via CDN isn't going to net much

Issue #994, opened 2016-06-28
Build easyCDN Module to handle CDN and Library Loading


mmilutinovic1313:
@heyMP and I talked a lot today about a more efficient way to add new libraries to ELMS and any Drupal site for that matter. I will be writing a custom module that will handle loading libraries and will allow us to choose where / how it is loaded, what version it will load, and if we want to free up library disk space. Attached is an initial drawup of @heyMP 's sketches of how we will structure the module. I will attach a TODO list as well in a further comment laying out my development plan and explaining the issue further.

![img_7730](https://cloud.githubusercontent.com/assets/7243665/16426314/760caa72-3d36-11e6-8acc-c40524136bb6.JPG)

btopro:
Interesting idea. Does seem like you should be able to use the libraries API info to snag / install libraries automatically, though that's file system permission dependent. I like the idea of remote loading them off a CDN by toggling on a big page.

mmilutinovic1313:
Description of what we want this module to do: 

The current implementation of implementing CDNs and Libraries API together on a system requires every component of the library to be set up properly. It needs to be detectable by `libraries_detect, proper versioning etc. 

The idea behind the easyCDN module is to have a remote system be able to allow developers to download new libraries, use old and new versions of libraries, and use CDN functionalities for libraries without having to rely on downloading libraries and modules that may or may not work properly with the Libraries API module. 

The modules will be checked out based on their searchable name on the JSDelivr CDN website. This is done by querying a search and getting a returned JSON object (http://api.jsdelivr.com/v1/jsdelivr/libraries?name=jquery). This object can then be searched for versions and all files contained within that version. This information will be the source of the info.json file that will be included in the easyCDN libraries directory allowing the user to have the ability to use different versions and search CDN's / local storage for files. If the user chooses to have this library downloaded locally - a new folder will be created in the easyCDN directory (either located in sites/all/libraries or another location that will allow for write access) with the version that was selected and the files that are contained in this version.

Doing it this way allows for multiple versions to be installed locally, but also allows users to remove old versions from the file system if they so choose. Developers can also just use the CDN to load the libraries which will free up space on the server. 

This module can make it easier for developers to test out new modules without needing to setup Libraries API friendly library structures manually.

mmilutinovic1313:
- [ ] Build App (will be broken into other tasks).
- [ ] Test easyCDN works on all ELMS:LN systems
- [ ] Verify that working on vanilla Drupal
[...]

PR #1437, opened 2016-11-02
#994 ELMS:LN RawGit implementation (#1234)

mmilutinovic1313:
We can load any library that we have locally via RawGit now. Also related to #1234


mmilutinovic1313:
This PR adds the rawgit support to elmsln only. Will be working on the ability to get rawgit from anywhere.

PR #1458, opened 2016-11-10
Allow HTTP OPTIONS request for anon users.

heyMP:
This will allow for preflight calls which take place in headless apps.

### DIFF
--- a/core/dslmcode/profiles/cle-7.x-2.x/modules/features/cle_critique/cle_critique.module
+++ b/core/dslmcode/profiles/cle-7.x-2.x/modules/features/cle_critique/cle_critique.module
@@ -50,11 +50,17 @@ function _cle_critique_api_random_critique() {
       $vars['submission'] = $submission;
     }
     else {
-      $vars = 'No Submissions available';
+      $vars = array(
+        'status' => '204',
+        'detial' => 'No Submissions available'
+      );
     }
   }
   else {
-    $vars = 'No Assignment specified';
+    $vars = array(
+      'status' => '400',
+      'detial' => 'No Assignment specified'
+    );
   }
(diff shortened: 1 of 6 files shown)


btopro:
why can't this happen in hook_boot? It you just use exit; instead of drupal_exit(); as that hasn't bootstrapped yet it should work / be a bit faster on response; just thinking outloud here.

btopro:
.... uhhh... how did this go from 5 lines to hundreds of files that have merge conflicts? we're on 0.6.x so you guys might need to pull / rebase / force pull

bradallenfisher:
we are working on a workflow... sorry.

heyMP:
I don't think hook_boot gives the cors module enough time to apply the headers.  It is doing it in hook_init. We could move it to hook_page_build so it doesn't happen so often, it will just increase the response time.  I'm fine with either.

btopro:
do we need the cors module? Coudl we just pull in what it's doing and hard code the exceptions?

btopro:
should I accept this or are you going to keep populating this one for a bit? fine w/ either

heyMP:
Yeah let's include the cors module for now.  We can always patch it or move it down the road.  I'm only using it for development right now.  I'll push up CORS.

btopro:
@heyMP if using your OPTIONS approach w/ CORS module then we'll need to ensure your stuff fires after the cors module sets it all in it's init hook. Right now it is but there's no real guarantee of it happening aside from C comes before E in module names :)

btopro:
@heyMP I added in a commit that makes sure elmsln_core runs later in its hook structure by making it load after everyone else. This should help ensure there's no race conditions between gaining control via alter hooks as well

PR #1505, opened 2016-11-22
fixes broken db from 7.51

### LINKED ISSUE #1486: install run and remove https://www.drupal.org/project/module_missing_message_fixer
Ran on new dev server against all sites now to see if it would work. It did. :)
However, since live at a version not ready for this module, every time we sync the db's down this will be undone. 

For now we will just have to ignore the drush errors.

bradallenfisher:
This will add the module_missing_message_fixer module. It will clean up a whole bunch of db tables that are no longer part of the schema.

drush command:
drush @elmsln mmmfl

https://www.drupal.org/node/2770115#comment-11750939
I have patched this so we don't need to dl the library for drush. it just works along side the module. 
Fixes #1486


bradallenfisher:
Once merged can you pull this into the prod server so I can run the drush command? After that there is really no need for this module, so I will add another pull request with the deletion.

btopro:
should we apply this via an bash update hook of some kind?

btopro:
needs to be semantic version name to match correctly. 0.6.0 also already exists as a version out there so needs to be higher then what currently exists; so, 0.6.1.sh as we'll cut a minor release update after getting lots of these little things worked out

btopro:
your killin it!

PR #1510, opened 2016-11-22
Studio changes into 0.7.x

heyMP:
These have a bunch of new changes that we need to get into 0.7.x


btopro:
Ok, few things:
- This should be against master so we can put it in 0.6.1 (just cause that's the next release slated for later today hopefully)
- take out the elmsln_core init hook thing bc that's already in

PR #1524, opened 2016-11-23
Added new assignments endpoints and assignments typecasting in angular.

heyMP:
Started stubbing out the endpoints for assignments.  I have some `@todo` notations beside functions to indicate that more functionality needs added.

Also, I created a little _cle_random() function to help generate mock content.  For instanstance, if there is a field that will be a list of options, I have the _cle_random() function generate different choices on page load to simulate a working endpoint until that field is actually hooked up.

screen shot 2016-11-23 at 2 39 27 pm


btopro:
Nice. Can you make some related issues for what needs filled in by me or brad

Issue #1436, opened 2016-11-02
create patch against leaflet / leaflet widget


btopro:
follow up from #479

There were a few changes made to leaflet and leaflet_widget modules to clean up the integration with ICOR. I think they are essential to making leaflet_widget usable in general since it resolves some conflicts w/ the latest version of leaflet.

@mmilutinovic1313 like the last patch related issue shadow @proconnor15 so he can get some more practice doing this / getting contributions back out there. PR back to our Repo when done for documentation purposes.

these live in `/var/www/elmsln/core/dslmcode/stacks/interact/site/all/modules/local_contrib`

mmilutinovic1313:
@btopro is @proconnor15 putting this patch as an issue on the leaflet drupal project?

btopro:
yes sir

btopro:
and leaflet_widget
they are 2 different things

proconnor15:
Submitted the patches on drupal and submitted PR.
https://www.drupal.org/node/2828094
https://www.drupal.org/node/2828085

proconnor15:
@btopro

btopro:
Excellent, these are exactly what they should be. Now put the patches in the related project folders and submit a PR here (not seeing the PR even though you said there is one)

PR #1544, opened 2016-11-30
Patches to leaflet and leaflet_widget module

proconnor15:
Committing patches to leaflet and leaflet_widget modules, related to issue #1436


btopro:
awesome, thanks man, we've been needing a lot of issue smashing / low hanging fruit wins to reduce the queue :)

PR #1546, opened 2016-11-30
support yaml and ldap

### LINKED ISSUE #1429: Add ldap module / ldap support
LDAP is very popular in most of EDU space as well as other large organizations. Let's add the LDAP module into the stack of globally supported modules and explore having a feature / extension off of elmsln_core that routes it into profiles (at minimum). Be nice to centralize / query LDAP if creds / module enabled so that we can present unified data about users.

### LINKED ISSUE #1533: php-pecl-yaml / yaml_parse_file not available outside cent6
this is needed to do Read the docs imports... which we never do so we've never noticed. It would be good to KNOW this will be available for when parse YAML stuff in the future.

#866

We'll want an upgrade hook for this

bradallenfisher:
This will fix the ubuntu side of things for yaml and ldap the rhel side was fixed in https://github.com/elmsln/php56-fpm-centos7-mysql56/commit/b85227575ba22c47ff50e973f7f8496a79885099

Fixes #1533
Fixes #1429

### DIFF
--- a/scripts/install/handsfree/ubuntu16/ubuntu16-install.sh
+++ b/scripts/install/handsfree/ubuntu16/ubuntu16-install.sh
@@ -46,7 +46,7 @@ apt-get -y install apache2
 apt-get -y install sendmail uuid uuid-runtime curl policycoreutils unzip patch git nano gcc make mcrypt
 
 #install php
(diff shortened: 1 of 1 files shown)


btopro:
I enjoy your company greatly.

Issue #1298, opened 2016-10-07
Media - attribute content to more than one course


rickhumphries:
Sometimes I have a piece of media content that I use across more than one course so I feel like I should be able to select more than one course here. 

![image](https://cloud.githubusercontent.com/assets/4973810/19195944/7cb87e22-8cab-11e6-832e-9fa02d775087.png)

Alternatively (and way cooler) would be if the system automatically detects that I've embedded that piece of media in a course, and tags it with each course it's been embedded in.

btopro:
hmmm. latest build is still single select. I do like the trackback idea though I'm not sure the best way to do that. In our legacy media system we have this happen at the media end point. So when media is loaded the nodes goes "dude who calls me.. oh, ok, let's stuff that here". Might add something like that in, not sure. good suggestion tho

btopro:
@mmilutinovic1313  Basically, just allow for multiple courses to be selected for all assets in media and interact

mmilutinovic1313:
Should be good to go pending PR merge

PR #1569, opened 2016-12-05
#1298 Support selecting multiple courses in elmsmedia & Interact

mmilutinovic1313:
Fixes #

Both now support multiple course selection. Update hooks added.


mmilutinovic1313:
Update hooks and multiple selection added. Let me know if it looks like I missed any. These were the only files in features that stated they had an override with the field change :)

btopro:
way too many files changed, doesn't seem like it was against 0.7.x. field_elmsln_competency for example should be dropped from the files changed

mmilutinovic1313:
Splitting this into two branches. To make sure no files are accidently kept.

PR #1691, opened 2017-01-07
Temporary 403 fixes for elmsln_api and restws

heyMP:


btopro:
rejecting since you told me to

PR #1714, opened 2017-01-11
i589 mooc flag feature

mmilutinovic1313:
Added mooc_flag feature to play with so we can push it to 7.x after testing some settings.


mmilutinovic1313:
Updated the permissions to not be default config

Issue #1747, opened 2017-01-17
Studio app feedback


btopro:
- [x] Make "home" navigation in the app website not the app loaded into Drupal
- [x] Make it more obvious that things go left to right for project listings; timeline style, pagination or something to suggest what's next
- [x] Privacy - does open me, to the entire class? I would consider Private (instead of closed) or Open
- [x] Critique Settings - Who sees that? I don't think students should set that.
- [x] Critique Style - Why is that separate? Can you combine that with the above option? Those options were not dependent on the above.
- [x] (fixed) We have 2 privacy settings currently, a toggle switch and a dropdown

btopro:
merging into requirements

PR #1757, opened 2017-01-18
Assignment form fixes

heyMP:
fixes for #1747 

Verify this access check looks correct: https://github.com/elmsln/elmsln/compare/0.7.x...heyMP:0.7.x?expand=1#diff-8b00181c3f7cb32e834ef7d47469c641L321


btopro:
Mobile so didn't check but yes that access check looks more accurate for future flexibility of that field

Issue #1615, opened 2016-12-14
review node revision delete module


btopro:
https://www.drupal.org/project/node_revision_delete

Good one for you @proconnor15 . This module would be VERY useful potentially as we keep growing to ensure we don't have any insanely large revision tables (doesn't hurt much but could).

Can you try installing this on a course space (you'll need a new VM bc of how much vagrant work we've done) and then make a ton of revisions to a page and ensure it's working / figure out how it works.

Ultimately if this DOES work we'd want it to be in every system with policies about what types to enforce it on as far as maximums. For example, we'd never have this on student work, but content pages, media assets, things in CIS, could have revisions for the last year and be reasonable to get rid of older stuff (if it has connotations like that).

btopro:
"delete any revision over 10, that are at least 3 months old"

btopro:
looks quite promising though it'll be a lot of settings that need set as it's per content type without the ability to set a default. I'd recommend doing this only in the courses, interact and media distributions, each with it's own feature per system that includes the settings for it.

proconnor15:
@mmilutinovic1313

proconnor15:
@btopro The module works how it says it would. The one thing is that it won't delete any of the revisions specified unless you run cron. Once you run cron all the revisions are deleted

btopro:
reviewed and added, longer term task to implement / start doing more real-world testing

PR #1799, opened 2017-01-25
#1615 node_revision_delete module added

proconnor15:
Added the node_revision_delete module to modules along with making sure you can delete revisions older than two years. Issue #1615


proconnor15:
Patch changes present in first commit

PR #1808, opened 2017-01-25
first stab at exporting angular lifecycle callbacks.

heyMP:
This is just a proof of concept and to get my math033 use case working.  I think we definitely need to flesh it out more.


btopro:
yeah we'll have to sit down and review w/ that lense since I dn't have a usecase for this yet but I know the media upload widget absolutely will need something like this

Issue #1817, opened 2017-01-27
Include revision_delete in all systems


btopro:
follow up from #1615. Like it says, get it worked into each distribution and tested heavily in MOOC and CIS to validate it for the rest.

proconnor15:
I'm working on implementing a minimum time that the revisions must be there before deleted. Should I just set the value to 2 years or should I make it so that you have the option to make it longer if you want?

btopro:
it should be a setting if the year / time value is when to delete stuff

btopro:
code is in place for this with a feature in mooc for testing though it's not implemented yet till it becomes an issue

PR #1824, opened 2017-01-30
Add node_revision_delete settings into mooc_cis_ux #1817

proconnor15:
#1817 added the dependency and the update function in mooc.install for the node revision delete module.


mmilutinovic1313:
Looks like a good start! Make sure you add the settings in your features export of the module. This will allow them to be set by default when the update script runs. 

Also @btopro there seem to be extra settings that were overridden? Do these look okay? Or do you know where they came from and how to remove them?

btopro:
stop using defaultconfig

btopro:
the menu links probably shouldnt have been dropped either

mmilutinovic1313:
@btopro that sounds good and the default-config is my fault. I didn't see the regular config menu in features export; only default-config and strongarm. Would we use strongarm instead of default-config then? Or just not add anything.

btopro:
strongarm

PR #1837, opened 2017-02-01
Added node_revision_delete to mooc_cis_ux try 2

proconnor15:
Added the node_revision_delete module to the dependencies and strongarm instead of to the default configurations. Continuation of pull request #1824
Issue: #1817


btopro:
looking good, make sure this does what it's supposed to by testing in a VM on your branch as next step

proconnor15:
@btopro  So what I see happening is that the time specified is when do you want to go delete the revisions when cron runs. The number of deletions per cron run is when it runs through each content type it can only delete x amount since it is a "database intensive task". To determine what to delete it goes to the content type with the oldest revision and will delete as many revisions until the number revisions you specified to maintain, or it will run out of the total number of revisions it can delete in a single cron run. After that content type it goes to the next oldest content type and does the same thing.

btopro:
@proconnor15 and it does this accurately without deleting things beyond the date it should correct? This is a pretty aggressive function so want to make sure there's a lot of testing on it before pulling it in.

proconnor15:
@btopro  No that is the problem. The time specified is only for how often it will check what revisions to delete. The only variable is how many revisions you want to keep. Not a certain date.

btopro:
oh... hmm... so like this would only run every 2 years basically if we say "2 years do x".

....hmm...

proconnor15:
Correct

btopro:
@proconnor15 this is a bigger issue then cause we'd want it to delete revisions after x number of years / time. we could set a half decent value that's reasonable (like 10) but still we NEED to keep it 2 years

proconnor15:
Yeah I have to look at it, I think it may work where when it runs every two years it just goes through and deletes any revision over the limit. Even if the revisions were more recent than two years. Which would be a BIG problem

btopro:
@proconnor15 yea, that's the edge case where it's problematic. This wouldn't be an issue... until 2 years down the road. It's still hypothetical but obligations to maintain records are what they are. There are other ways of handling this (like forcibly taking back ups once a year or something).

I'm not opposed to the every 2 years this thing runs idea but let's explore putting in that safety net to not delete unless older then x

proconnor15:
Agreed. I can look at what they do, and see if I can find a way to input a length of time before the records can be removed

btopro:
cool, closing this for now then cause we don't want ot accept until the underlying module supports what we talked about

proconnor15:
Sounds good

Issue #1848, opened 2017-02-02
Illegal String Offset 'fid' in CIS when course created with Syllabi / Welcome Letter


mmilutinovic1313:
Got this error when creating a new course with Syllabi today. I feel like I've seen this before somewhere, but checked through the issue queue and couldn't find reference to it. Here's the screenshot below: 

![illegal string offset fid](https://cloud.githubusercontent.com/assets/7243665/22559355/d39abc4c-e93e-11e6-994c-3d4262b58611.png)

Looks like the files are not coming across properly and the `fid` is not being set. Here is the snippet of code it errors at: 

```
// make sure the file came across
      if ($form_state['values'][$file_name] != '' && $form_state['values'][$file_name]['fid'] != 0) {
```

mmilutinovic1313:
`$file_check = array('syllabus', 'welcome_letter');` is where the `$file_name` variable comes from. It has no child elements of its own and therefore `fid` cannot be found.

mmilutinovic1313:
@btopro Just changed it to ` if ($form_state['values'][$file_name] != ''` because `fid` doesn't exist anyways: Putting PR in shortly

PR #1849, opened 2017-02-02
#1848 Removed `fid` check for file_check as fid does not exis

mmilutinovic1313:
This removes error caused in #1848. GitHub is adding your extra commits again...but only one file is changed so all should be good.


btopro:
does this actually fix the issue (aka syllabus file does save correctly and is associated) or does this just make the error go away?

mmilutinovic1313:
It tells me that I have to be outside of the master section to view it. I can see it in CIS. But I'm still not sure how to switch out of Master section in a course. Student view?

mmilutinovic1313:
Creating a course with no section data to test...

btopro:
unrelated issue, that would resolve itself after a nightly sync up (for the sections)
can't send section data at a course that doesn't yet exist. What we could do is after a system is built it could call home and say "what sections I got?". These "additional settings" aren't things we use currently so this is edge-casey to begin with as people add offerings to get more offerings into the system

mmilutinovic1313:
That's fair - Well this does remove the message and the syllabus shows up on CIS for the course. So I'd say this fixes the issue because if the files don't exist it won't show up in the form state and therfore the if statement will resolve to false and not do a file_load.

btopro:
that's all I needed to know ;)

btopro:
@mmilutinovic1313 if u pull latest there's also a fix I pushed today that helps make sense when viewing master on a syllabus page as to how to resolve.

mmilutinovic1313:
Sounds good - I'm putting 2 screenshots in here as reference :)

pdf saved

they saved

Issue #1853, opened 2017-02-02
Documentation Request: PC Environment Instructions


mmilutinovic1313:
I think we should add instructions for PC install support. Whether this is "Download a Unix Command Line Tool" or setup instructions themselves. I'm testing out the steps presented in this article (http://www.howtogeek.com/249966/how-to-install-and-use-the-linux-bash-shell-on-windows-10/) as an example for getting a PC environment to use Unix commands.

ernharthsw:
![windows command error](https://cloud.githubusercontent.com/assets/25517075/22571403/991815a4-e96d-11e6-80ae-bd33d32bfb9e.JPG)

rickhumphries:
I use putty.exe as my command line tool on a PC (http://www.putty.org/)

In fact, I could probably write this guide if you like?

mmilutinovic1313:
@rickhumphries that would be great!

btopro:
cygwin

rickhumphries:
Just looked at cygwin..think I got the wrong end of the stick. My guide would pretty much be, get a digital ocean server, download putty, use that to log into server,  how to change host file.. etc. Can do that if that's what you meant but anything else is beyond me I reckon

btopro:
lol

btopro:
Vagrant works on PowerShell if you run as an admin too btw

mmilutinovic1313:
I'll figure this out once I pull up my Windows partition

btopro:
#1868 doesnt have a ton of implications for this but worth mentioning that the install routine for vagrant has been simplified

mmilutinovic1313:
I was shown that the git download listed on the install file can be used to run unix commands and should fix the command problems for windows users.

mmilutinovic1313:
Closing as merged

PR #1876, opened 2017-02-15
#1853 Windows Clarification

mmilutinovic1313:
Windows users can use git download shell for unix commands if they do not have built in unix support.

Fixes #


mmilutinovic1313:
#1853

PR #1878, opened 2017-02-16
Fix Apache config files to work around letsencrypt bug

MarcoMoreno:
Certbot's letsencrypt fails to put whitespace between `</VirtualHost>` and `</IfModule>` which confuses Apache.  This causes the installer to revert changes and remove all SSL virtual host files.

See https://github.com/certbot/certbot/issues/2267

(This is my first ever pull request so hopefully I'm doing it right! :)


btopro:
so what is this PR doing cause I don't see any changes if you look at the files claiming to be changed

btopro:
also is it only an issue in these 3 files? there's lots of conf files

MarcoMoreno:
This is only a problem if the `</VirtualHost>` does not have a newline following it.  All the other conf files have a `<DocumentRoot>` that follows so they don't need to change.

btopro:
that works for me. congrats and thanks for the contribution!

btopro:
@MarcoMoreno in the future, please open an issue in the queue, then submit the PR so we have some flow of problem identification to resolution

PR #1904, opened 2017-02-24
Revision_deletion added

proconnor15:
Related to #1817 
Added the revision_deletion module instead of the node_revision_delete module.  Revision_deletion looks at the minimum age of a revision before it deletes it. The deletions are run through cron automatically. (Or can be done manually) 
Then created a feature included the settings to delete them when cron runs and after two years.  Also added two patches. One for the addition of the option to delete after two years. The other was a patch someone submitted to Drupal which fixed a major error, but has not yet been approved


btopro:
Move the stuff currently in mooc_cis_ux over to the mooc_revision_deletion feature you made. All this stuff should be self contained so that eventually when we enable mooc_revision_deletion by default we'll have a very simple update hook / dependencies[] = to add

proconnor15:
@btopro I believe I had not removed the strongarm stuff for the node_revision_delete. I think I removed that now. Although the dependencies for the revision_deletion module should be in the mooc_revision_deletion

btopro:
finally got a chance to review, nice work. We won't activate it yet but good to know it's all in place when we need to solve this.

PR #1973, opened 2017-05-09
Automatic lockout after 5 attempts - admin controls #1965

### LINKED ISSUE #1965: Automatic lockout after 5 attempts - admin controls
Just a feature suggestion really. We've had a few students receiving the 'sorry you've been locked out because of 5 failed login attempts' because they can't enter their passwords correctly.

Thought it might be good to have some admin controls for unblocking users etc. with these modules:
https://www.drupal.org/project/flood_unblock
https://www.drupal.org/project/flood_control

cgldevel:
Fixes #1965 

We've added two modules that should help w/ automatic logout; we haven't enabled them though.

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/modules/ulmus/flood_control/LICENSE.txt
+++ b/core/dslmcode/shared/drupal-7.x/modules/ulmus/flood_control/LICENSE.txt
@@ -1,0 +1,274 @@
+GNU GENERAL PUBLIC LICENSE
+
+              Version 2, June 1991
+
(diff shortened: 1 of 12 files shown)


btopro:
@cgldevel please fill out the "fixes #" part of the PR so we know what this is adding specifically. Tagged to the correct issue in the name so that's good but we need to know that this is adding the modules to the code base but not actually implementing them at this time.

btopro:
And with that, @cgldevel has had a successful PR to building the future. Let's print out this commit message and frame it. Welcome to the revolution.

cgldevel:
@btopro stoked to be part of it! :)

PR #1988, opened 2017-05-12
tweaks to mooc toolbar.

heyMP:
Request to bump to the font-size down a hair to allow more room for the page titles.

## Before
screen shot 2017-05-12 at 9 16 38 am

## After
screen shot 2017-05-12 at 9 16 26 am


cgldevel:
I think it looks nice slightly smaller as well.

PR #1995, opened 2017-05-13
Added support for gitbook in auto redirection.

heyMP:


btopro:
uhh... what is field_git_description and where did it come from? I don't think that's in normal gitbook is it?

heyMP:
I assume it was in the gitbook feature, I didn't put it in there.  It's probably safe to assume that students should not ever see a gitbook page.  I'll make a change.

heyMP:
I'm moving this to its own branch.

Issue #2005, opened 2017-05-16
Mooc nav icons overlay text


heyMP:
Bug:
![screen shot 2017-05-16 at 9 48 03 am](https://cloud.githubusercontent.com/assets/3428964/26109382/707021fe-3a1d-11e7-8ee0-411bb27d2246.png)

PR #2006, opened 2017-05-16
Mooc nav icons overlay text #2005

heyMP:
fixes: #2005 

screen shot 2017-05-16 at 9 47 51 am


btopro:
wow, this is bizaro world. you pushing commits to break designed elements that I made / accounted for?

PR #2066, opened 2017-06-23
open studio and recent comments updated

bterp-web:
styled recent comments - done

continued styling open studio


btopro:
Looking good, maybe make a comment that's really long and play w/ the truncated / fading away aspect of the mock up a bit.

PR #2083, opened 2017-07-03
WIP:  Refactor roster processing for high volume

### LINKED ISSUE #1945: Refactor populate_course_network cis command to batch users
Issue:
Very high enrollment courses / sections (1,500+) are currently causing issues finishing the sync in larger builds. This makes sense cause it's attempting to load, verify and resave / add that many people in a single pass, then move onto the next system.

resolution:
We assemble the roster per course and already do batching in this way to distribute load but we need to also put a cap on the number of users we'll process at a time. This would allow for batching within the context of a listing of course users. To do this, we either ship all the users and have each site only process 200 at a time before re-posting to itself, OR we do the batching ahead of time and let CIS broker this out.

alternate resolution:
Let's see if we can leverage a crazier part of httprl to multi-thread function processing. If we dropped into a multi-threaded approach to solving this by crunching the same function with MORE workers (sounds bad I know) that each then have their own process that has the ability to not timeout, we could actually have the intial thread keep moving forward while a non-block loop keeps churning through the current info. This was designed to keep long jobs from timing out.

Examples from httprl:

```

// List of nodes to load; 241-244.
$nodes = array(241 => '', 242 => '', 243 => '', 244 => '');
foreach ($nodes as $nid => &$node) {
  // Setup callback options array.
  $callback_options = array(
    array(
      'function' => 'node_load',
      'return' => &$node,
      // Setup options array.
      'options' => array(
        'domain_connections' => 2, // Only use 2 threads for this request.
      ),
    ),
    $nid,
  );
  // Queue up the request.
  httprl_queue_background_callback($callback_options);
}
// Execute request.
httprl_send_request();

// Echo what was returned.
echo httprl_pr($nodes);
?>



Load nodes 50-100 using httprl_batch_callback and node_load_multiple.

<?php
// List of nodes to load; 50-100.
$nids = range(50, 100);
// Run not parallel if background callbacks are disabled.
if (!httprl_is_background_callback_capable()) {
  $results = node_load_multiple($nids);
}
else {
  // Queue & Execute requests.
  $results = httprl_batch_callback('node_load_multiple', $nids);
}
// Echo what was returned.
echo httprl_pr($results);
?>



Load nodes 50-100 using httprl_batch_callback and node_load.

<?php
// List of nodes to load; 50-100.
$nids = range(50, 100);
// Run not parallel if background callbacks are disabled.
if (!httprl_is_background_callback_capable()) {
  // httprl_run_multiple does a foreach on $nids running every $value through
  // the given callback.
  $results = httprl_run_multiple('node_load', $nids);
}
else {
  // Set options.
  $options = array(
    'multiple_helper' => TRUE,
  );
  // Queue & Execute requests.
  $results = httprl_batch_callback('node_load', $nids, $options);
}
// Echo what was returned.
echo httprl_pr($results);
?>


Load nodes 50-100 using httprl_batch_callback and node_load_multiple as user 1.

<?php
// List of nodes to load; 50-100.
$nids = range(50, 100);
// Run not parallel if background callbacks are disabled.
if (!httprl_is_background_callback_capable()) {
  // Run node_load_multiple as user 1
  $current_account = $GLOBALS['user']->uid;
  $GLOBALS['user'] = user_load(1);
  $results = node_load_multiple($nids);
  // Set global user back.
  $GLOBALS['user'] = $current_account;
}
else {
  // Set options.
  $options = array(
    'context' => array(
      'uid' => 1,
    ),
  );
  // Queue & Execute requests.
  $results = httprl_batch_callback('node_load_multiple', $nids, $options);
}
// Echo what was returned.
echo httprl_pr($results);
?>



Run a function in the background. Notice that there is no return.

<?php
// Call watchdog('httprl-test', 'background watchdog call done');
// in the background/async.
httprl_call_user_func_array_async('watchdog', array('httprl-test', 'background watchdog call done 2'));

?>
```

andrewbelcher:
Fixes #1945 

- [ ] Test thoroughly
- [x] Enable processing of the queue via Drupal's cron
- There are a couple bits where I wasn't sure if the current behaviour was a bug, so may want to revisit once I have clarity on that.
  - [x] Should the accounts to be archived be per section or global (currently global)?
  - [x] Decide whether storing the role cache is too much data - need an idea of how many roles we are likely to have... I'm inclined towards it would be fine.
  - [x] Make the archiving of students more granular - have held off until I know whether the list needs to be global or not.
- [x] Switch calls to `_cis_section_create_accounts()` that are suitable for queue processing to use the queue.


andrewbelcher:
So far I have refactored the roster processing into a class that is capable of storing the state across multiple requests. By default (and every call) it will process all immediately, but if a batch size is provided, it will process up the that number of items and then queue itself.

Still to do:
- Test thoroughly
- Enable processing of the queue via Drupal's cron
- There are a couple bits where I wasn't sure if the current behaviour was a bug, so may want to revisit once I have clarity on that.
- Switch calls to `_cis_section_create_accounts()` that are suitable for queue processing to use the queue.

andrewbelcher:
So this includes processing via the cron and updating `cis_course_authority_elmsln_api_callback_sync_data_model()` and `cis_course_service_elmsln_api_callback_sync_data_model()` to use the deferred method. I am pretty sure looking at the code that these are both called as part of the site sync process, so should be safe to defer.

The remaining use is in `_cis_section_activate_section()`, I don't know whether this usage is safe for deferral?

andrewbelcher:
There's also `_cis_section_sync_roster()` which I think is unclear whether it can be deferred...

andrewbelcher:
Have made mail/pass sync on existing accounts and made cron based processing an opt-in via a variable for BC.

btopro:
Archive is global to that site, ppl should only ever be in 1 section

btopro:
Per role amount there are like 8 or 9 total, max would be like 12

andrewbelcher:
@btopro is the following scenario not currently possible:

- Section 1:
  - *had*: User 1
  - *has*: User 1, User 2
  - *deactivate*: Nothing
- Section 2:
  - *had*: User 2, User 3
  - *has*: User 3
  - *deactivate*: Nothing , as User 2 is in our global active list

So having a global active list means that if a user moves to a section earlier in the roster, they will not be removed from the latter section. This doesn't seem right, or is moving between sections impossible?

btopro:
umm... yeah your right that scenario is theoretically possible and has not been accounted for. It also hasn't been something we've even thought of because it's not really a thing in either of our scenarios but your right it is possible someone would drop from a section and switch to another. Edge case but possible.

andrewbelcher:
My suggestion would be to make the active students track per section, which will resolve the edge case. I can't think of any potential negatives to it...

btopro:
yeah me neither; be good to account for it if your there doing it anyway

andrewbelcher:
Ah, should have read the code more carefully, archive doesn't remove section specific stuff, so global is definitely right.

andrewbelcher:
Ok, so I have switch to defer by default, if configured globally. I've made it so each student archive counts as a process (i.e. broken it up in case there are lots).

From my perspective, it is finished pending testing. I will try and do some tomorrow!

rickhumphries:
@btopro I'm playing around with v0.9.x as it currently is with our settings (custom CSS, legacy Zurb Foundation switched on) and it's too big of a jump for us at the moment . v0.8.x (master) is OK though - so if it's ok could we commit these changes against master when the time comes? That would allow us to implement this particular feature sooner and only have to make the jump to v0.8.x at the same time. 

That'll give us a better window for figuring out how our custom styles are going to work in a web components era. Thanks

btopro:
Andrew will need to rebase these changes against master so that it doesn't pull in like 300+ changes from the theme layer. I wouldn't expect a version release for at least 2 months, we're going to be basically radio silence until after Aug 20th working on the studio

rickhumphries:
Sure, np. @andrewbelcher are you OK to rebase this against master please?

andrewbelcher:
@rickhumphries sure, I can do that. I'll include the other relevant commit as part of it. I'll try and get that done today.

rickhumphries:
Thanks @andrewbelcher 👍

andrewbelcher:
@rickhumphries #2118 is the PR for master.

PR #2089, opened 2017-07-05
Add Gitbook sync feature 2009

### LINKED ISSUE #2088: deprecated: Clean up permissions after running git-book drush commands
Bug:
- after running git-book-create or git-book-sync the file permissions are changed from apache to the current user.

Need a way of running permissions clean up after drush commands, or running drush command as apache user.

heyMP:
Fixes #2088

Added a gitbook sync feature.  You can run `git-book-sync [name_of_book]` and it will pull changes from origin and run the parser to either create or updated content in the drupal book.

I also added to git config changes when creating the repo to minimize erroneous merge conflicts:
```
$repo->run('config core.fileMode false');
$repo->run('config core.autocrlf false');
```

There is one outstanding issue that I'm not sure how to solve, #2088.  However, if you manually clean up permissions with `leafy` after running the drush commands it works fine.

### DIFF
--- a/core/dslmcode/stacks/courses/sites/all/modules/local_contrib/git_book/README.md
+++ b/core/dslmcode/stacks/courses/sites/all/modules/local_contrib/git_book/README.md
@@ -28,6 +28,50 @@ Take that zip file, and put it's contents into a folder called "gitphp" in your 
 Also, you need to make sure your Drupal site has write access to your private directory. This is where it will store the
 git repository.
 
+## Giving Apache the ability to push to gitbook remote repos
+
+If you are going to enable the `push` option in your gitbook, you need to make sure that apache can make push new commits to your
+remote repo. To do that you need to create an ssh key for apache.
+
+### Create an ssh key for apache user
+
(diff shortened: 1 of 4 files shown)


heyMP:
This has been tested in production.

btopro:
https://github.com/elmsln/elmsln/pull/2089/files#diff-6e41d5185a604b148e2a434572af6e0c

btopro:
@heyMP ensure that this isn't added as a submodule but instead the files

heyMP:
whoops. that wasn't supposed to be in that branch.  Sorry about that.

btopro:
before accepting this, Reason me this: If the content is backed by on file system version control, do we actually need to track revisions in the database? right now this commit (mirroring the previous settings) has `$node->revision = 1;` could we not disable the DB bloat and just have these changes tracked in the file system repo?

heyMP:
Yeah, good point.  Right now, the revision system is the only way for a content author to revert since we don't have the GitBook app working yet.  So I guess two options:

- Remove revisions for now anticipating that we'll have a way for content authors to revert via a polymer git app
- Keep revisions until we build out the git

btopro:
good point, added in for now

Issue #2014, opened 2017-05-17
Studio UX Mockups


michael-collins:
I'm just getting started on these. I'm thinking that a vertical nav is a good idea to separate it from the rest of the horizontal navigation we have going on. This is just a photoshop mockup until we can get the entire ux nailed down. I'll update as I make progress.

<img width="1267" alt="screen shot 2017-05-17 at 2 26 07 pm" src="https://cloud.githubusercontent.com/assets/1085304/26169876/3d2500e4-3b0d-11e7-95c3-793ffadbc812.png">

michael-collins:
![image](https://cloud.githubusercontent.com/assets/1085304/26170611/11cd1cbc-3b10-11e7-9e75-67de4773e85d.png)

michael-collins:
<img width="1260" alt="screen shot 2017-05-17 at 2 49 01 pm" src="https://cloud.githubusercontent.com/assets/1085304/26170635/1d24f134-3b10-11e7-87e6-3105136c17f0.png">

btopro:
I wonder what a left panel would look like if shaded to match the thing you are working on or how color could help reinforce that. google+ does this to help discern between different aspects of google plus once in that "app" so to speak -- https://plus.google.com/collections/followed

btopro:
what is a discussion feed? Like what people are saying about stuff centric instead of submission centric?

michael-collins:
Instead of the comments tab, I renamed it discussions. I haven't had a chance to put nearly as much time in today as i wanted. Spent all day reworking a tutorial for one of their assignments, and building stuff out  on elms.

btopro:
gotcha, ok, just clarifying

btopro:
also did u make the above icons the the mock up? If so, can I have them at 30px svg?

michael-collins:
Those are material, trying to standardize on them as much as possible.

<img width="1216" alt="screen shot 2017-05-17 at 11 57 07 pm" src="https://cloud.githubusercontent.com/assets/1085304/26186012/970edefa-3b5c-11e7-9958-8a4eff0a69e3.png">

michael-collins:
Let's work color in at a later time. We should focus on greyscale for now, and introduce color after we have things working a bit better. I'm not anti color, it's just harder to design with, and takes a more wholistic overview to do it right.

[...]

PR #2094, opened 2017-07-07
lrnapp-open-studio / lrndesign-gallerycard

bterp-web:
screen shot 2017-07-07 at 1 51 11 pm

screen shot 2017-07-07 at 1 51 45 pm


btopro:
This is looking awesome :) Minor thing, for the future  try to tag commits / a PR with a related issue so we can track it back some place if needed. This can be traced back to #2014

PR #2118, opened 2017-08-04
Issue #1945 by andrewbelcher: Switch to Drupal's queue for roster syncing

### LINKED ISSUE #1945: Refactor populate_course_network cis command to batch users

andrewbelcher:
Fixes #1945 for master.

See #2083 for 0.9.x PR.


btopro:
@rickhumphries were you able to use this successfully yet before I merge it?

rickhumphries:
@btopro I know Andrew was able to test this successfully but I haven't tested this on our development server yet. I wasn't sure if there was anything extra that was required for this to work in terms of how the crons are scheduled on the course sites first? Is that what came out of the Skype call when Andrew was demoing it?

btopro:
Ya cron on things not the CIS will be need to be told to run every 15 minutes or so to ensure that they all process the things they are supposed to.

rickhumphries:
OK, is that something we setup to be specific to our system or something that's setup in core? Sorry if that's a stupid question

btopro:
this would need configured manually by someone modifying crontab, like ur new hire :)

rickhumphries:
So would this need to be run every 15 mins by every non-CIS site simultaneously every 15 mins? Or would they run one after another? Would this get called from the CIS itself or does each stack have its own crontab? ...if you or @andrewbelcher could guide me in the right direction for what we need to do from this point forwards I'd be grateful. 

I guess it could still be merged at this point anyway, as there's still an option to use the previous sync method

btopro:
I'd prefer to wait till we have confirmation fro real world data and deploy that it doesn't have adverse affects.

as for crons you can call things like `drush @courses-all cron --y` to have it execute cron on all the courses. If you are using other systems then you can call it against them to as separate cron operations. I'm still not positive crontab is the most efficient way of getting the jobs to run through vs having the cron queue call itself again to restart its execution thread (so like it keeps eating its own tail basically to chew through how ever many students). If the queue call when it realized it needed to stop then sent another call to itself and shut down it would achieve the same affect but without needing crontab management; @andrewbelcher does that sound reasonable? We send non-blocking call structure with httprl all the time in these type of calls.

andrewbelcher:
The Drupal core cron will already allow it to 'eat its own tail', but with some time limits around it to prevent it just running indefinitely and impacting other aspects of the site. 

For many reasons, you should have the core cron running on all your sites anyway. There are things to do with cache, maintenance tasks etc. So if you've got 200 sites, you're going to need a server and setup that is going to be able to manage that anyway. I think this demonstrates an intrinsic scalability problem of each course operating as it's own (multiple) site... 

Having sad that, the core cron is not the only way we can utilise the queue. My suggestion would be we add a drush command that can run that specific queue which will run to completion, with some locking to prevent multiple concurrent runs. Your crontab can then execute this regularly for all the relevant sites, which means you don't need to core cron running as frequently. 

I'm swamped the rest of this week, but I should hopefully be able to find some time next week to write that for you.

btopro:
I will agree with you on any other drupal site. We don't run cron's regularly by design since the multi-milti-site setup is effectively for letting people get their own little compartmentalized clone of a space to work with. Every space works the same (more or less) and the diff sites are for user / content segregation. So usually with sync routines the sync spider hits one site, when done, it tells the site next to it to run. So there are very long running jobs, usually nightly, to get everything seeded. This works as well since courses can be accessed 24/7 but honestly people pop in for a few minutes here and there.

Anyway, a drush thing that can run through the queues would be ideal. That's something we could kick off a background job for effectively and while sync would seed things a job on the backend could ensure the rest of any queues are chewed through every x number of minutes or whatever.

andrewbelcher:
@rickhumphries @btopro Here is an update that moves the CLI processing of the roster from the cron into a separate drush command - `drush cis-process-roster`. This will run until complete.

It makes use of locking, so you can use a crontab to trigger it regularly, but once it has started, it will run to completion without another process hitting at the same time.

However, one thing I noticed, at least in the vagrant setup, is that the `lock_acquire` is broken on CLI. Anything that uses `lock_acquire` from CLI will be provided a lock even if another process has already claimed it.

rickhumphries:
That's great. Thanks for this @andrewbelcher . So were you able to process the full list of 10,000 fake students by just running the cron once? Let me know if this looks OK @btopro

btopro:
@rickhumphries sounds like from just running that drush function once

btopro:
@andrewbelcher Looks like a minor merge conflict related to this being updated to support passing around user data object for SIS / internal properties to set on user objects. Should be minimal tweak to your function to match
https://github.com/elmsln/elmsln/blob/0.9.x/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/features/cis_section/cis_section.module#L495
```
// support for data object
        if (isset($data)) {
          $fields['data'] = $data;
        }
```
type of thing

andrewbelcher:
@btopro I think that sorts it - we don't need to do anything further with creating a new user as we already use all of `$user_data`.

btopro:
@andrewbelcher nice. Can you add the 'data' array info in via the `$edit` optional second variable for user_save? This way it'll merge what it found in the DB w/ what came across so it's meshing with as opposed to overwritting this field

https://api.drupal.org/api/drupal/modules!user!user.module/function/user_save/7.x

btopro:
also adding field_name to your og_group call; we had to ensure it was very specific bc of a data model issue as opposed to OG just guessing which field we meant.
```
$values = array(
        'entity_type' => 'user',
        'entity' => $account,
        'field_name' => 'og_user_node1',
      );
      og_group('node', $gid, $values);
```

andrewbelcher:
Updated...

btopro:
looks good!

PR #2123, opened 2017-08-09
Removed global css selectors from applying to polymer elements.

bterp-web:
This makes it so that global css does not effect polymer elements.

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/themes/elmsln_contrib/foundation_access/css/main.css
+++ b/core/dslmcode/shared/drupal-7.x/themes/elmsln_contrib/foundation_access/css/main.css
@@ -1,43 +1,43 @@
 @charset "UTF-8";
-*:not(.style-scope),
-*:before:not(.style-scope),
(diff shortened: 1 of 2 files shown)


heyMP:
Looks like it put the style scopes within the clip property.

Example:
```
clip: rect(1px:not(.style-scope), 1px:not(.style-scope), 1px:not(.style-scope), 1px);
```

We'll have to remove those instances.

bterp-web:
Got it. It was also added to any rgba

`box-shadow: 0 2px 5px 0 rgba(0:not(.style-scope), 0:not(.style-scope), 0:not(.style-scope), 0.16):not(.style-scope), 0 2px 10px 0 rgba(0:not(.style-scope), 0:not(.style-scope), 0:not(.style-scope), 0.12);`

btopro:
It's a good idea even if a bit aggressive. Let's start doing this with anything that we find coming into conflict and make sure all global CSS going forward is better scoped

PR #2127, opened 2017-08-09
Studio File Fields

heyMP:
screen shot 2017-08-09 at 4 45 51 pm


heyMP:
Also, fixes the links field.

heyMP:
Also, it's worth noting that I made change to the `_elmsln_api_v1_file_output` function to set a default value for `file['display']`

PR #2131, opened 2017-08-13
Submission type

heyMP:
Added submission_type to cle_submission content type.

Submission will check its parent assignment to see what type of submission it should be. We might want to add a presave condition to assignments to prevent them from changing critique method if there are already submissions started.

If we are returning content from our endpoint then we have to return it as 200 or 201 or around there.  Browsers won't parse the body of a 500 error.

I added a toast to the submission-button that will let the user know if there are no submissions to critique at this time.  You can turn off the toast by setting `display-errors` to `false` on the submission-button.


btopro:
still no edit mode but less broken then yesterday for me

heyMP:
Try to create a new submission and see if that one works.  Might be broken for submissions that previously been created before this pull request.

btopro:
If something doesn't have the field type it should default it to returnING a normal submission for consistency of legacy material

heyMP:
Yeah that makes sense.

Issue #2302, opened 2017-10-18
HTML comments in wysiwyg cut off content further down


rickhumphries:
Bug:
- v0.8.1

- Our content outline is full of HTML comments where we signal to beginning/end of different style templates/components. It looks like wherever a comment is inserted - the content below this doesn't show on the wysiwyg

**View on the page:**
![image](https://user-images.githubusercontent.com/4973810/31716532-34a6b400-b400-11e7-9ac1-fb1e863e2ce9.png)

**WYSIWYG**
![image](https://user-images.githubusercontent.com/4973810/31716565-60495982-b400-11e7-9822-6731c2ae2c67.png)

**Source Code**
![image](https://user-images.githubusercontent.com/4973810/31716597-87e68bd6-b400-11e7-8618-53ddea1c59f7.png)

btopro:
that's weird.... have you tried placing the comment after the closing </p> there. Also maybe see if the "correct faulty HTML" option is in the input filters `admin/config/content/formats/textbook_editor`

It's on by default and might be misidentifying this text and "correcting" it

rickhumphries:
Placing a comment after the closing...? I've switched off that option in the input filters but it didn't have an affect. 

I also deleted all the comments from the source to make sure that it was definitely them causing it, and deleting them makes everything appear in the wysiwyg - so pretty certain it's that.

btopro:
the comment is within a <p> tag; what if it's by itself like not in a tag does that help at all?

rickhumphries:
No that doesn't seem to do it. I'm trying to figure out a pattern to it but can't...It's definitely trying to process it in some way though, even with the 'correct faulty HTML' option switched off

btopro:
does it display while in CKEditor?

rickhumphries:
No,  so the WYSIWYG view show above is the view I get in CKEditor. 

It's almost as if the CKEditor view isn't recognising the closing comment tag or something so is just cutting it off.

I also notice that if I keep switching between source and view mode in CKEditor, it keeps adding paragraph tags after the closing comment -->

rickhumphries:
...but it only does that with some comments, not others. urgh

btopro:
there's some additional wysiwyg clean up in 0.9.0 for webcomponents so it doesn't do that weird <p> tag addition thing u mention. This happens when ckeditor doesn't know what the tag is (it's behavior that seems intentional for some reason).

[...]

PR #2305, opened 2017-10-23
Include HTML Comments in CKEditor Protected Source

JackWillDavis:
Related to issue #2302

Fix for an issue where HTML comments entered into CKEditor Source View would cut off content when switching back to WYSIWYG mode. Includes HTML comments and keeps previous support for WebComponents.


btopro:
tested and looks like it's working on my end locally. Welcome to the revolution @JackWillDavis

PR #2314, opened 2017-10-28
Update h5p.md

yurg:
H5P has a module for Drupal 7 & 8 now.

Fixes #


btopro:
Thank you for the PR, I'll update this manually though as it's completely out of date. Because of updates in the 0.9.0 release you no longer have to install H5P manually and so we don't need any of the language associated with this

btopro:
https://github.com/elmsln/elmsln/blob/master/docs/enduser/media/h5p.md newer version, removed that entire section; thanks for pointing out this was out of date though!

PR #2333, opened 2017-11-19
resolves #2332 for Nikki to review

### LINKED ISSUE #2332: Align text the correct way in all scenarios
By eliminating choice, we can ensure users always align text the correct direction.

btopro:
based on suggestions from @nikkimk this should enable all of our text to be aligned correctly in all scenarios.


heyMP:
Looks really aggressive and I'm not sure what this would solve.  I would have to see the use case.

btopro:
All text should be centered in all possible scenarios and it should be very difficult to un-center. As @nikkimk always says:
previewinstancedata

nikkimk:
Um... let's talk about this. I agree with @heyMP, and I may have some edge cases that we should link about.

Issue #2168, opened 2017-08-22
Create pandoc microservice


btopro:
#1590 and recent #2163 have illustrated that our PDF printer is a weak link. It's also largely unsupported (the service powering it). This means we should really take a step back and explore crating a micro-service based on pandoc in order to do document conversion. This would also allow us to import and export as other formats like word docs, markdown, pdf, etc. It also means it would be easier to do things like make a pdf of a person's project in studio, currently not possible via dompdf (the library powering our PDf printing).

btopro:
https://github.com/elmsln/elmsln/tree/0.10.x/core/microservices/pandoc files in place in 0.10.x but need routines to boot it up as well as pathways towards secure usage

btopro:
this is actively being worked on by a student team which will go into HAX directly

PR #2351, opened 2017-12-08
Removed dompdf dependencies on mooc_ux_defaults

JackWillDavis:
## Fixes ##

We get a 500 error when attempting to use a different print_pdf module library (i.e. wkhtmltopdf) when generating PDFs. Issue can be fixed by removing the dompdf_library helper module and removing the dependencies from mooc_ux_defaults.

## Commit Message ##

Removed dependencies on mooc_ux_defaults which was causing a 500 error when attempting to use a different library (i.e. wkhtmltopdf) for generating PDFs.


btopro:
hmm.. we'll need something else in this that uses an install hook and maybe has a global setting you guys can implement to make it not enable. I get why you need to disable it but for people that DO use it it's autoconfigured out of the box to provide at least SOME solution until we rearchitect the whole thing to be pandoc delivered (at some point)

btopro:
once #2168 is implemented via #2384 then we can accept this while at the same time removing all the stuff that provided it :)

btopro:
@JackWillDavis changed my mind as we just had the DOMPDF stuff start bloating server calls bc of how much processing a single one was taking. only real way around this is to implement #2168 on a lambda or something

JackWillDavis:
Yeah, went with WKHTMLTOPDF this end as it’s a little less bloaty. Lambda function might be better option long run though

btopro:
yeah, the fact that it's affecting page performance to generate a PDF of something is kinda silly

PR #2359, opened 2017-12-27
ELMSMedia dashboard with pagination.

heyMP:
This is the first shot at a paginated dashboard for the media system using hax-source.

screen shot 2017-12-27 at 3 27 16 pm

Still needs a lot of work but if you wanted to check it out I'm submitting via a pull request.

This does have a breaking change in hax-source.  The search feature is no longer a default, you have to specify it with a 'search' attribute.  I also added an 'auto' attribute to auto-fetch results

Still needs:
 
- [x] Url updating when selecting pages and filters
- [x] filters / tags
- [x] search
- [ ] make card clickable
- [ ] sort criteria
- [ ] add more info to card


btopro:
Skim on phone, this I flipping awesome Starting point. I think we'll need variability on what invokes thr paged result to load so we can target youtube and others in a flexible manner.

core/webcomponents/services/ELMSMediaService.php
```diff
@@ -0,0 +1,175 @@
...
+      $encoded_item->relationships->group->data->id = $item->og_group_ref[LANGUAGE_NONE][0]['target_id'];
+      // author
+      $encoded_item->relationships->author = new stdClass();
+      $encoded_item->relationships->author->data = new stdClass();
+      $encoded_item->relationships->author->data->type = 'user';
+      $encoded_item->relationships->author->data->id = $item->uid;
+      $encoded_item->relationships->author->data->name = $item->name;
+      // // Actions
+      // $encoded_item->actions = null;
+      drupal_alter('cle_open_studio_app_encode_project', $encoded_item);
```
btopro: assume left over from copy and paste

core/webcomponents/services/ELMSMediaService.php
```diff
@@ -0,0 +1,175 @@
...
+   * Get a list of projects
+   */
+  public function getMedia(array $options = NULL) {
+    $return = array();
+    $query = new EntityFieldQuery();
+    $query->entityCondition('entity_type', 'node');
+    /**
+     * @todo this should be configurable
+     */
+    $query->entityCondition('bundle', array('elmsmedia_image', 'external_video', 'svg', 'h5p_content', 'document', 'video', 'audio'));
```
btopro: could probably do a not clause and then just ensure that cis_course and section don't get returned since they are data model based; everything else should be up for grabs. This most likely can be abstracted to your other class so that you can query from any of the systems via our REST engine any (generally) get the right response so we can start visualizing everything :) (gizmo type return would need some clean up / expansion obviously unless we added something into the sites themselves to say what types they are)

heyMP:
Added some filtering functionality

screen shot 2017-12-29 at 2 35 06 pm


@btopro I haven't gone through your comments yet.

btopro:
merge whenever you feel like; this is looking great and really laying the foundation for decoupling Drupal that much more in each of our micro-services via the backend service classes your making.

heyMP:
Added Search feature.

screen shot 2018-01-04 at 2 04 11 pm

Issue #2357, opened 2017-12-19
Updates to Discuss Templates


rickhumphries:
v0.8.1

I've been asked to set up a forum on one of the courses we offer so just testing this out. It looks like I need to change some of the template styling though to account for our lack of user profile photos.

![image](https://user-images.githubusercontent.com/4973810/34166620-451cf0e4-e4d7-11e7-9453-53d42f991684.png)

Ideally I'd like to just add the username of the author in here instead. What's the best way of going about this? I imagine this is probably going to be replaced by something built in web components at some point so just kind of looking for a short term solution.

btopro:
yeah, this was mostly just positioned correctly and has a decent data model but wasn't finished. It's templates are driven by the Harmony project which we use some of the data model of https://github.com/elmsln/elmsln/tree/master/core/dslmcode/profiles/discuss-7.x-1.x has the distro. 

https://github.com/elmsln/elmsln/tree/master/core/dslmcode/stacks/discuss/sites/all/modules/local_contrib - has those projects.

I THINK that those come from views templating from those modules but I'm not positive. Hope this is enough of a push in the right direction.

rickhumphries:
Thanks - we'll dig around and see what we can do

JackWillDavis:
Hi @btopro, we've managed to resolve by amending the template directly but is there a way of protecting this from future updates? For example, is there somewhere in the ELMS config files where we can define our own harmony templates to use by default?

Or will it have to be done via a patch file?

btopro:
well what's the patch look like? We don't actively use the discuss system at this time. It's very low in our priority though it needs a make over (and will get one ala webcomponents at some point). Guess what I'm saying is depending on what it is if it's a generic solution we can accept it into core :)

JackWillDavis:
It is basically commenting out the code which shows the profile image and places the username in there instead. Very minimal tweaks really.

btopro:
Oh ya let's do it then, it's a bandaid anyway. Have you checked out the studio in a local development environment to see what's available in 0.9.0 and the 0.10.x branch? It's much more refined then the discussion boards currently

On Jan 3, 2018 7:07 AM, Jack Davis <notifications@github.com> wrote:

It is basically commenting out the code which shows the profile image and places the username in there instead. Very minimal tweaks really.

btopro:
PR accepted; issue fixed

PR #2365, opened 2018-01-04
Harmony Profile Image adjustment

JackWillDavis:
This PR contains a tweak to the harmony discussion core which 

Removing the profile image from discussion posts and replacing it with the username of the poster. Default link added to make it match in with local styling.

It definitely a bit hacky, adding a new user_load now I look at it again but if there is a better way of grabbing the username of the poster then please let me know :)

### DIFF
--- a/core/dslmcode/stacks/discuss/sites/all/modules/local_contrib/harmony_core/theme/harmony-post.tpl.php
+++ b/core/dslmcode/stacks/discuss/sites/all/modules/local_contrib/harmony_core/theme/harmony-post.tpl.php
@@ -45,7 +45,7 @@ $links = render($content['links']);
   </div>
   <div class="clearfix">
     <div class="post-left post-user-profile text-center">
-      <a href="#"><?php print $username; ?></a>
+      <a href="#" tabindex="-1"><?php print $username; ?></a>
     </div>
     <div class="post-content post-left-offset">
       <?php print render($content); ?>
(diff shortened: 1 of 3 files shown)


JackWillDavis:
As discussed in issue #2357 with @btopro and @rickhumphries

btopro:
@JackWillDavis can you also apply it to the file in question and commit that? I appreciate you have the patch here but we're good with accepting this change as well so you don't have to run patched (patch helps us track in the event we ever upgrade that aspect)

JackWillDavis:
OK, I'll just make the change now

JackWillDavis:
All done!

btopro:
so this will work in your setup but in 0.10.x we should use the function that helps resolve names as it's cleaner and allows for integration w/ remote services / alternate name providing services like SIS -- `_elmsln_core_get_user_name()`

btopro:
https://github.com/elmsln/elmsln/blob/0.10.x/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/features/elmsln_core/elmsln_core.module#L2300 -- this will figure out the name to provide with a fallback of ->name

JackWillDavis:
OK, updated and neatened up a bit. Let me know if this is OK. I've updated the patch file but not sure if you wanted changes to files AND a patch file but it's all there now.

btopro:
nice! is there a reason for wrapping the name in a link tag?

JackWillDavis:
It was so it didn't need to be restyled in CSS. If it's wrapped in a link, it keeps the same design for the username as used elsewhere in the post template.

btopro:
ahhh ok. can you through a `tabindex="-1"` on that to improve accessibility then since it's just purely for styling

JackWillDavis:
I'll remember that for the future :) Added now.

btopro:
for future A11Y reference - that tabindex tells the screenreader / you tabbing through on the keyboard, to ignore that link. This is useful when things that would otherwise get focus shouldn't get them for whatever reason

JackWillDavis:
I think I remember using tabindex years ago for ignoring links on an advertising slider but I'll bear accessibility in mind for PRs here.

PR #2376, opened 2018-01-10
Add Print Outline Button

### LINKED ISSUE #2375: Add Print Outline Button
Feature Request:
- A button under "alternative formats" that prints the entire outline, not just the page.

screen shot 2018-01-10 at 4 08 47 pm

heyMP:
Fixes #2375


btopro:
this can crush sites if not cached as it'll try and load everything; especially in media intense courses making lots of calls out; we had this functionality probably 3 years ago and removed it for this reason; however, we're much more optimized then we were back then at all levels

heyMP:
Most likely a better solution would be to have an on-demand, pre-built PDF of the outline.  Possibly one that is rebuilt nightly if the content has been updated?  More of how gitbook does it.

btopro:
v true

PR #2415, opened 2018-02-11
Integrate Web components module with Markdown module

### LINKED ISSUE #2412: Markdown filter wrapping web components with erroneous <p>
Bug:
If the Markdown filter is presented with a web component it will wrap the element with <p> tags.  That doesn't sound bad until you have html content within the web component slot.  If you do, then the markdown filter will attempt to "fix" the html structure and break your dom hierarchy.

### No html in slot

screen shot 2018-02-10 at 12 10 44 pm

screen shot 2018-02-10 at 12 10 55 pm

### HTML in slot

screen shot 2018-02-10 at 12 11 10 pm

screen shot 2018-02-10 at 12 11 18 pm

heyMP:
When Markdown module checks to see what 'custom' tags it should ignore, we jump in with `_webcomponents_all_tags()` to load every web component that it knows about and adds it to the list of protected or "clean" tags.

I included a patch file.  I'm not sure what our naming convention is so I just named it by issue number.

Fixes #2412


btopro:
glad that works, @heyMP patch file is empty btw

PR #2469, opened 2018-05-25
docs: fix possessive "it's"

apetro:
Fix instances of "it's" that were meant to be "its"


btopro:
This fixes it in many comments documented in code which I won't keep track of. Drupal code policy is to patch files and include the patch directly with it

apetro:
I think you're trying to tell me that some of the files in `elmsln/elmsln` aren't locally maintained they're copy-and-pasted from some kind of upstream, am I getting this right?

btopro:
ya, how bout we just do MAINTAINERS, GOVERNANCE since ppl who like spelling will see those. The rest of the comments are mostly upstreams from d.o. so not too concerned about spelling there :)

PR #2589, opened 2019-01-18
Added a raw flag to elmsln_section tokens.

### LINKED ISSUE #2588: Add a raw option for elmsln_section tokens
Feature request:
elmsln_section tokens rendered in the page have an outline indicator for to show admins which content is dynamically rendered to the page.  However, some users utilize the tokens within html.  If we add an option for cis tokens to indicate that it should not include the indicator markup and just render the raw output of the token.

related: #2508

heyMP:
You can add the `:raw` to an elmsln_section token and it won't add the outline indicator to it.

screen shot 2019-01-17 at 10 59 40 pm

screen shot 2019-01-17 at 10 59 28 pm

Fixes #2588


btopro:
Does the _data variation of the current tokens not have this ability? Been awhile since I added that but would want to know the difference between the two

btopro:
Never mind I see. Raw let's it pass through and not be evaluated. Good call

PR #2595, opened 2019-02-09
added alter hook for adding custom network or "More Apps" items to th…

### LINKED ISSUE #2500: Adding Network Links to 'More Apps' Side Menu
v0.10.x

I'm currently looking at an upgrade path from 0.8 to 0.10 and one of the things I'm not sure about it how to add a menu link to the 'More Apps' flyout menu using the edit network links page `/admin/config/user-interface/cis-lmsless-nav`

If I add the group as 'Network' it shows as an icon in the `etb-network-nav` bar just fine, but previously we were able to set the group as something custom like 'Company Links' and it would appear in it's own section here:

image

heyMP:
Fixes #2500

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.api.php
+++ b/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.api.php
@@ -7,10 +7,34 @@
  * top CIS "LMSless" bar. Because this is a module that provides the area
  * and "themes" it, it's harder to get at the underlying variables via
  * preprocess functions.
+ * 
+ * This exectutes AFTER some of the menus are assembled through render()
+ * 
  * @param  array &$vars theme variables for rendering to the top bar.
  */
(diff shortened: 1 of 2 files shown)


btopro:
also _cis_connector_build_registry is what feeds this, which has it's own hooks cis_service_registry / _alter. These are for registering applications worth passing credentials around about / recognizing as parts of the network topology.

This is said for reference purposes, what you're adding is for theme / "I put a link to make it easier for people" perspective which is perfectly fine.

heyMP:
Just so I understand would this code work?  From looking through the code, I couldn't track what `$vars['services']` is doing.

```
function hook_cis_lmsless_theme_vars_alter(&$vars) {
  $vars['services']['External']['youtube'] = array(
    'title' => t('YouTube'),
    'url' => 'http://youtube.com',
    'machine_name' => 'youtube',
  );
}
```

btopro:
services show up in the fly out, the other method mentioned is for the network bar in the top left that students see. There's some obnoxious reasons to do both vars and I can't recall what it is right now. Has to do w/ template flexibility and the fact that you could tell something to show up in both if you really wanted to this way (which now that I think of it WAS the reasoning) vs if it was just the same var collection then you couldn't.

PR #2596, opened 2019-02-12
User Block Theme changes

JackWillDavis:
Switched the `theme()` and `drupal_alter()` function calls in `cis_lmsless.module` to allow changing of the user_block *before* it gets processed by the theme.

### Suggested Change
At the moment, the `user_block` is defined before `drupal_alter('cis_lmsless_theme_vars', $vars);` is called which means that devs are not able to make $vars changes take effect in the user flyout menu on the front.

Proposing running the alter hook before the theme is run.


btopro:
makes perfect sense. Thank you!

PR #2597, opened 2019-02-12
Control over the user flyout login/profile links

JackWillDavis:
### Description

We've had the need in the past to edit the label and URL of the profile links in the user flyout menu. Thought it might be useful for others as well.

Proposing switching the `$userprofile` and `$userlink` to arrays so that they can be amended easier within `hook_cis_lmsless_theme_vars_alter` without the need of a patch file.

### Commit Message

Switched `$userlink` and `$userprofile` to arrays so they can be changed in `hook_cis_lmsless_theme_vars_alter`


JackWillDavis:
Hang on, needs a quick fix

JackWillDavis:
Done

btopro:
this seems reasonable and more flexible then my lazy fix to make it an lrnsys button in the first place :0

PR #2603, opened 2019-02-21
"My Profile" button for Anonymous users in flyout menu

JackWillDavis:
Currently the "My Profile" link tries to appear for Anonymous users in the flyout menu, despite no data being available.

The `cis-lmsless-template.tpl.php` template is currently checking that `<?php if (!empty($username)) : ?>` however `$username` automatically defaults to "Anonymous" [here](https://github.com/elmsln/elmsln/blob/0.11.x/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.module#L331) which results in a blank button (as in the screenshot below).

my profile link

This PR just adds a second to check to make sure the button isn't rendered for Anonymous users.


btopro:
may want to wrap that Anonymous in a t() function. I have a feeling this can be overwritten / language specific. other then that sounds reasonable

JackWillDavis:
Ahh good point! I’ll update it tomorrow morning.

JackWillDavis:
I've updated with the `t()` function now

btopro:
looks good, thanks for the fix

PR #2610, opened 2019-03-05
Mooc Nav Block Hook

JackWillDavis:
Added a new hook allowing modules to alter the menu tree used in the book navigation block

#### Description

Added a new hook (`hook_mooc_nav_block_alter`) which allows us to alter the menu tree used to build the navigation block on book pages. This is being used internally to develop a new feature where we need greater control over the links displayed in the sub navigation but could be useful for others.


JackWillDavis:
Not sure if you have a naming convention for core hooks so I just used the name of the module but let me know you wanted it renamed.

btopro:
that should be ok... though it might also target `mooc_nav` since there's a block named that. maybe (silly I know) make it `mooc_nav_mooc_nav_block`

JackWillDavis:
Sorry just realised I never replied to this. I've updated the name of the hook and added a second one for allowing the same thing on the other book navigation.

Issue #2652, opened 2019-09-01
Remove Sponsors from README


btopro:
remove sponsors we dont use that anymore

McGarvelcuddy:
I removed the sponsors, and created a Pull Request.

PR #2660, opened 2019-09-09
Removed backers and sponsors

RCS-167:
Fixes #


btopro:
thank you @RCS-167 , this fixes #2652

PR #2661, opened 2019-09-09
Removed Sponsors

McGarvelcuddy:
Removed sponsor section at the request of btopro

Fixes #


btopro:
Thank you @McGarvelcuddy github protip - if your commit or message above referenced #2652 then it would automatically create a notice showing the two are related (in the notification to the maintainer as well).

PR #2680, opened 2020-10-22
Pagination fix

JackWillDavis:
Fixes an issue with pagination where an empty result set could result in a negative `$start` value.

image


JackWillDavis:
I'll change this to a patch file but will let you have a look first and you can let me know it's better fixed elsewhere in the codebase

btopro:
Looks super minimal to me. Save the patch file in the views folder that the patch would be applied from. Drupal Core is the only place I deviate from this method by making a big PATCHES folder so I don't accidentally lose / forget to apply things (that and I have to check change records to see if they were fixed, etc). so put it in top level of the views folder and should be g2g

JackWillDavis:
Changed them to patches. Also created a patch to get around an undefined index issue which was popping up during installation.

Would you normally apply these patches in the repo as well or just drop the patches into the folders?

btopro:
These look reasonable to apply and they are in a good location. Roll another commit in that applies these and then I think we're good to accept this :)

JackWillDavis:
Patches applied!

PR #2681, opened 2020-10-23
Ubuntu 18.04 Installer updates

### LINKED ISSUE #2626: Ubuntu 18 handsfree installer
Hi all,

Is a handsfree installer for Ubuntu 18 on the roadmap for the future?

As a partly related question: At the moment, we're moving onto PHP7. Everything looks like it's working on a fresh copy of ELMS but is there anything we should be aware of? If you've had any issues at all of course.

JackWillDavis:
Fixes #2626 

These are all of the changes I needed to make to get a complete installation working on Ubuntu 18.04:

* I've made changes to the `cis_connector.module` file as it was throwing a couple of errors - just using `parse_url()` to break up the URL which seemed to solve the problem in this case
* Applied a fix to the `editableviews` module (and included the patch) to stop a breaking error from appearing
* Adjusted the symlink for the CIS profile
* Edited the `elmsln-install.sh` file to:
* * Apply a `dr-locations` option to the site-install command for authorities as it was struggling to find the recipes for CIS for some reason
* * Ensured that the `admin` role has the `administrator` role for CIS
* * Manually set the `install_profile` variable to the dist value - should be done automatically by installing the profile but was bugging out for CIS in my case
* Added PHP7.2 service restart calls to `handsfree-install.sh` and `elmsln-preinstall.sh`
* Bumped up the `mcrypt` module version to 1.0.3 for Ubuntu 18.04 and added support for non-interactive install as it was stopping to ask for a confirmation
* Rearranged the IF statements in `elmsln-preinstall.sh` to add support for Ubuntu 18 installations - now checks for 18, then 16 and treats any other Ubuntu installation as 14 or below (wasn't sure how far back you wanted to support on that)

As we were talking about in #2626 it might have been those `views` errors were messing up the CIS install so some of these extra drush calls might not be necessary now (they're being addressed in #2680 at the mo) but I haven't tested it without them as of yet.


btopro:
so you were able to repair the installer with this??

JackWillDavis:
Yeah, it's running through smoothly now on a fresh installation. I'm not sure how much the views module was causing the problems but working now.

btopro:
This is excellent news and thank you for all your hard work tracking this down I know it wasn't easy! Views has been problematic w. These minor notices in php the last few php minor versions

PR #2683, opened 2020-11-16
ddev fixes and updates

JackWillDavis:
Including a set of fixes to ELMSLN ddev support in this PR:

* Added ddev HTTPS support
* ELMSLN now passes the ddev username into the container which means folders can be correctly assigned by using `$DDEV_USER`
* Adding extra mysql conf command to improve database response times
* Dockerfile now includes `cron` in the web container automatically
* Ensured that cron is now working correctly (even on a restart)
* There is now a restart script which helps restore drush/job-file functionality after a restart
* ELMSLN should stop installing if it detects the container already has an installation
* Removing old apache conf files
* Including a README file in the ddev folder with instructions

Also had an issue with drush failing when restarting the web container. This was due to the routine cleanup of some home folder files being deleted by ddev. Fixed this by:

* Automatically including `.bashrc` and `.bash_alises` files which include all of the ELMSLN config
* Including the ELMSLN drush alias folder into the home directory by default. Drush is rebuilt in the home folder on every spin up.
* Including a blank `~/.my.cnf` to prevent ddev from automatically creating one which was preventing drush from bootstrapping the database.

The cron support is a little wobbly which I might need a bit of help with from someone (it might be I've been staring at it for too long now!). It'll work a lot of the time but not consistently. Might just be a ddev/docker issue and not something we can work around but I've included a manual command in the README which can be run if the sites aren't building from jobs files automatically.


btopro:
Fantastic work  or "brill" (did I British right?)

JackWillDavis:
Haha spot on! Cheers!

PR #2691, opened 2021-09-09
Bakery php7 replace mcrypt

JackWillDavis:
This PR fixes the `bakery` module which won't work in PHP versions above 7.1 due to dropped `mcrypt` functionality. 

Edited the Bakery module itself and added a patch file which lines up with ELMSLN's pre-patched `bakery.module` file.


btopro:
excellent; thank you for tracking this down!

PR #2692, opened 2021-09-10
Adding in Mooc hooks

JackWillDavis:
Adding in a few hooks to allow devs to amend the course navigation before its rendered.


btopro:
`mooc_nav_mooc_nav_block` is walking dangerously close to something that could have a naming conflict bc it ends in block meaning that `mooc_nav_block_alter` which is a core hook would be triggered if not for writing `mooc_nav` twice :). However I get why you did this and the others are fine

JackWillDavis:
Haha yeah, I remember us talking about this when I was setting up module locking a while back and that was the hook name we decided on. Happy to change it though if you wanted, just be a small tweak in our own repo for the hook.

btopro:
nah it's fine; just wanted to plant that thought for future hook additions :)

PR #2693, opened 2021-09-10
Updating the `harmony-post.tpl` file.

JackWillDavis:
Editing the HTML structure of `harmony-post.tpl`


JackWillDavis:
Ah yeah! I've just added a patch file now

PR #2694, opened 2021-09-10
Tincan API Bridge - Allow devs to prevent sends.

JackWillDavis:
Allow modules to prevent xAPI posts via the `tincanapi_ajax_data_alter` hook. 

We have a particular use case for this where we prevent certain verbs from actually being sent off to the LRS to cut down on the amount of data being transferred/processed there.

Tweaked it slightly from what we were using in our patch so let me know if it looks fine your end.


btopro:
So if I understand correctly this has "hey send this" default true, then code you've written can alter that away. So the default behavior is the same yes? I think this makes sense just confirming

JackWillDavis:
Yeah, similar to how `drupal_mail` works where you can hook in and alter `send` to be false to prevent the email being sent. Defaults to being sent normally though yep

PR #2695, opened 2021-09-10
Allow entity iframes to display messages.

JackWillDavis:
Fix to allow status messages to be printed within embedded iframe content. Can be useful in some cases, such as embedded webforms, to be able to see if there are any system messages being displayed.

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/entity_iframe/entity-iframe.tpl.php
+++ b/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/entity_iframe/entity-iframe.tpl.php
@@ -17,6 +17,11 @@
  * @ingroup themeable
  */
 ?>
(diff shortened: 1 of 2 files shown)


btopro:
I could be wrong but I don't know that messages will EVER be false and will always write the HTML into the iframe. Are you able to confirm that this is actually NOT going to print anything if there are no messages?

JackWillDavis:
I can confirm that it doesn't for us at least. I don't think `$messages` would be false as such but it would be empty/blank which would be treated as false wouldn't it?

We have messages showing up for certain conditions, here's the HTML for when there is no message printed on the iframe webform:

image

And this is what's shown when we ARE printing extra messages for the learner:

image

btopro:
as long as it wont influence the height then its fine. I just worry about blank space when we're rendering a video or whatever thing in context of another page and itll auto resize to the height of the item

PR #2696, opened 2021-09-10
Switch section context

JackWillDavis:
Thought I would throw this one up as well while I was at it lol!

This allows anyone who's been given the `switch section context` permission to be able to switch sections (instead of limiting it just to staff/teacher roles), whilst keeping the masquerade feature limited to the original staff/teacher groupings. This allows admin to give certain roles the ability to move between sections without giving them the additional permissions associated with teachers/staff.

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.module
+++ b/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.module
@@ -571,7 +571,7 @@ function _cis_lmsless_theme_vars() {
         ),
       );
     }
-    if (_cis_connector_role_groupings(array('staff','teacher')) || isset($_SESSION['masquerading'])) {
+    if (_cis_connector_role_groupings(array('staff','teacher')) || user_access('switch section context') || isset($_SESSION['masquerading'])) {
       // account for systems without sections (like authorities)
       if ($section) {
         // ensure they can see the section switcher link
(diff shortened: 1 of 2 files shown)


core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.module
```diff
@@ -584,7 +584,7 @@ function _cis_lmsless_theme_vars() {
       if (isset($_SESSION['masquerading'])) {
         $vars['masquerade_logout'] = '<lrnsys-button label="' . t('revert (@name)', array('@name' => $GLOBALS['user']->name)) . '" href="' . url('masquerade/unswitch', array('query' => array('destination' => current_path(), 'token' => drupal_get_token('masquerade/unswitch')))) . '" icon="supervisor-account" hover-class="' . $vars['lmsless_classes'][$vars['distro']]['color'] . ' ' . $vars['lmsless_classes'][$vars['distro']]['dark'] . ' white-text"></lrnsys-button>';
       }
-      else {
+      else if(_cis_connector_role_groupings(array('staff','teacher'))) {
```
btopro, line 587: doesn't this then remove the ability to "un-masquerade" if this is in an elseif?

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.module
```diff
@@ -571,7 +571,7 @@ function _cis_lmsless_theme_vars() {
         ),
       );
     }
-    if (_cis_connector_role_groupings(array('staff','teacher')) || isset($_SESSION['masquerading'])) {
+    if (_cis_connector_role_groupings(array('staff','teacher')) || user_access('switch section context') || isset($_SESSION['masquerading'])) {
```
btopro, line 574: this makes sense

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_lmsless/cis_lmsless.module
JackWillDavis: Nah, the un-masquerade button is in the if/isset section which runs before this. If the user is already masquerading, it won't check whether they're a student/teacher role group but I think that's how it worked before anyway.

So, if the user is already masquerading they get the "Revert" button from Line 585. If they're not already masquerading, and not in the staff/teacher role group, they won't be shown the form to masquerade at all.

JackWillDavis:
Forgot another tweak which I've just added in which amends the sections being displayed.

PR #2700, opened 2021-11-05
Import URI fix for reactive-element

JackWillDavis:
I've just updated our development server to v1.2.0 and I think there is an issue with the built version of lit-element. It's trying to load in `reactive-element` but it's using the wrong URI and resulting in a 404 error.

image

Issue is in this file:

`/var/www/elmsln/core/dslmcode/cores/haxcms-1/build/es6/node_modules/lit-element/lit-element.js`

I've tweaked the import from:

```
import{ReactiveElement as e}from"./node_modules/@lit/reactive-element/reactive-element.js";
export*from"./node_modules/@lit/reactive-element/reactive-element.js";
import{render as t,noChange as l}from"./node_modules/lit-html/lit-html.js";
export*from"./node_modules/lit-html/lit-html.js";
```

to this:

```
import{ReactiveElement as e}from"../lit/node_modules/@lit/reactive-element/reactive-element.js";
export*from"../lit/node_modules/@lit/reactive-element/reactive-element.js";
import{render as t,noChange as l}from"./node_modules/lit-html/lit-html.js";
export*from"./node_modules/lit-html/lit-html.js";
```

Which has fixed the error.


btopro:
thank you for the report. I know what this is though this is a machine built file so I can't accept this PR. Running through our tooling to try and correct this. Issue ends up being that multiple things call for slightly different versions of `lit` and then it installs in multiple directories, which can generate this issue. For example, a node_modules directory shouldn't be in another node_modules directory (until I screw something up :) ).

doing a local build now to see if this issue is still there. Your fine to roll this fix for now but future builds will undo the fix

PR #2701, opened 2021-12-09
Added new hook for adding new options to courses

JackWillDavis:
Added a `hook_mooc_content_render_options_alter` hook to allow us to manually add new options to the course subnav.

Currently we're using this hook to load in a custom region where we can add new blocks containing additional buttons for the course subnav, like this:

```
function modulename_mooc_content_render_options_alter(&$content){
  $blocks = block_get_blocks_by_region('operations');
  $content .= drupal_render($blocks);
}
```

image


btopro:
Looks good. I'll ask in slack about this more

PR #2702, opened 2021-12-17
Updating ckeditor_custom_config.js

JackWillDavis:
Updates to CKEditor settings to be more compatible with HAX authoring including allowing inline elements to wrap block elements and switching off auto-formatting

This PR includes fixes which prevents the issues we had with CKEditor adding loads of extra `p` and `span` tags in content and prevents it from doing weird formatting of custom webcomponents.

Related to: https://github.com/elmsln/issues/issues/915

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/themes/elmsln_contrib/foundation_access/ckeditor_custom_config.js
+++ b/core/dslmcode/shared/drupal-7.x/themes/elmsln_contrib/foundation_access/ckeditor_custom_config.js
@@ -17,8 +17,11 @@ CKEDITOR.editorConfig = function( config ) {
   // this helps ensure there's no conflict with made up elements
   // and automatical <p> wrapping
   config.autoParagraph = false;
-  // protected source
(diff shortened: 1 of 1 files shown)


btopro:
`protectedSource` was put in place to ensure that webcomponents didn't get stripped. Does this mean that everything else is kept the same and now it just doesn't mess w/ the output? We're obviously moving away from CKEditor but still have similar use-cases to you where there's a setting keeping us tied to the form in certain instances.

JackWillDavis:
Yeah, protectedSource seemed to keep the bug alive for whatever reason. With this PR, we don't get any webcomponents stripped out and everything stays the same without CKEditor doing it's own thing.

If you want, I could add that line back in? I tried loads of regex changes for protectedSource but couldn't find one which stopped the issue whereas taking it out completely sorted it out.

btopro:
I trust you won't mess up your own stuff even more out of spite :). I think it was there for `<template>` tags and others to not process but we don't do as much with those and honestly I'm beginning to wonder if that protectedSource was a hold out from V0 spec days of HTML imports.

JackWillDavis:
Nice one thanks. We definitely used to rely on it for v0.x but seems not to be needed now. I'll keep an eye on it and if we need to add it back in for any reason, I'll make another PR :)

PR #2703, opened 2022-02-07
Mooc content hooks

JackWillDavis:
@btopro 

Just adding in the code block you sent over a few weeks back for fixing loading issues related to multiple shortcodes on one page. This was the related issue: https://github.com/elmsln/issues/issues/878

Let me know if you needed anything else!


btopro:
looks good; glad this helped

PR #2704, opened 2022-02-18
Tweaking cmstags endpoints structure

JackWillDavis:
Just switching up the cmstags endpoints into an array format so they can be easily overridden in the `webcomponents_cmstokenendpoint_alter` hook.

Also removed the token endpoint being manually added in `_webcomponents_cmstags_process` as this overwrites any hook which edited that endpoint. I've not noticed any issues after removing that my end but let me know if it's important your end.

I can see the webcomponents_cmstokenendpoint_alter hook is being used in core which will need editing but I can't tell if it's required any more: https://github.com/elmsln/elmsln/blob/837cb1bb19c15137ef052e105c7ed5e66eb3c05d/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/cis_connector/modules/core/cis_service_connection/cis_service_connection.module#L180

### DIFF
--- a/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/webcomponents/modules/webcomponents_cmstags/webcomponents_cmstags.module
+++ b/core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/webcomponents/modules/webcomponents_cmstags/webcomponents_cmstags.module
@@ -167,6 +167,8 @@ function _webcomponents_cmstags_token_render($format) {
     'editEndpoint' => $url,
     'editText' => t('Edit this content'),
   );
+  // allow developers to override the content
+  drupal_alter('webcomponents_cmstags_response', $return);
   // output the response as json
   print drupal_json_output($return);
   exit;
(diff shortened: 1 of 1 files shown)


JackWillDavis:
Adding a new hook to allow devs to edit anything in the `$return` response array from webcomponents_cmstags

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/webcomponents/modules/webcomponents_cmstags/webcomponents_cmstags.module
```diff
@@ -318,16 +323,6 @@ function _webcomponents_cmstags_tips($filter, $format, $long = FALSE) {
  * Callback function to perform the content processing.
  */
 function _webcomponents_cmstags_process($text, $filter, $format, $langcode, $cache, $cache_id) {
-  // ensure we only apply this once based on the input format
```
btopro, line 321: why is this block removed?

btopro:
`_webcomponents_cmstags_process` is going to run even if this is remote rendered and injected into content. I'm not positive if that block is needed but would need to do some real testing to ensure it's not a problem.

core/dslmcode/shared/drupal-7.x/modules/elmsln_contrib/webcomponents/modules/webcomponents_cmstags/webcomponents_cmstags.module
JackWillDavis: So, this was removed as it seems to override whatever changes are made to the `window.cmstokenEndPoint` in the `webcomponents_cmstokenendpoint` hook.

A dev could use the `hook_webcomponents_cmstokenendpoint_alter` hook to update the cmstokenEndPoint value to something else. However, this block would just set it back to the default regardless.

I did just try removing this block entirely and it started using the correct endpoint I'd set in the hook and I haven't seen any issues crop up because of it (at least on our setup).

PR #2706, opened 2022-12-23
chore: update jsonwebtoken to version 9

ChristianMurphy:
Fixes security warnings raised by dependabot for this dependency.
More details on the issues in the changelog.

https://github.com/auth0/node-jsonwebtoken/blob/master/CHANGELOG.md#900---2022-12-21 https://github.com/auth0/node-jsonwebtoken/wiki/Migration-Notes:-v8-to-v9

this also removes and ignores the yarn error log, which are for local debugging only


btopro:
thank you @ChristianMurphy it is appreciated!

